PR #5, opened 2019-11-04
Chafey/mem outfile

chafey:
This PR includes several things:
* mem_outfile - enables encoding to a memory buffer
* raw_in - enables encoding from a raw pixel source.  This shares a bit of code with yuv_in so there is opportunity to refactor commonality
* support for capturing resolution and level truncation points during encoding to support future decoding partial bitstreams
* function cpp_init_j2c_encoder in ojph_wrapper.cpp which uses mem_outfile and raw_in to encode a raw pixel buffer.  I also update the test code in the BUILD_TEST ifdef which requires files on my machine not committed (this code wasn't working anyway though)

Also - sorry for not squashing the commits, will do that next time


src/core/others/ojph_file.cpp
```diff
@@ -191,4 +198,92 @@ namespace ojph {
...
+    assert(this->size);
+    assert(this->data);
+    assert(this->cur_ptr);
+    
+    this->is_open = false;
+    this->size = 0;
+    this->data = NULL;
+    this->cur_ptr = NULL;
+  } 
+
```
aous72: You forgot to deallocate memory

src/core/common/ojph_file.h
```diff
@@ -81,6 +84,48 @@ namespace ojph {
...
+    // release ownership of compressed data bytes
+    void release_data();
+     
+  private:
+    bool is_open;
+    size_t size;
+    ui8 *data;
+    ui8 *cur_ptr;
+  };
+
```
aous72: This class has been added to the code.

src/core/common/ojph_file.h
```diff
@@ -56,6 +56,7 @@ namespace ojph {
     virtual size_t write(const void *ptr, size_t size) = 0;
     virtual void flush() {}
     virtual void close() {}
+    virtual size_t tell() = 0;
   };
 
   ////////////////////////////////////////////////////////////////////////////
```
aous72: Added

src/core/common/ojph_file.h
```diff
@@ -56,6 +56,7 @@ namespace ojph {
     virtual size_t write(const void *ptr, size_t size) = 0;
     virtual void flush() {}
     virtual void close() {}
+    virtual size_t tell() = 0;
```
aous72: Added to code

src/core/common/ojph_file.h
```diff
@@ -72,6 +73,8 @@ namespace ojph {
     OJPH_EXPORT
     virtual size_t write(const void *ptr, size_t size);
     OJPH_EXPORT
+    virtual size_t tell();
+    OJPH_EXPORT
```
aous72: Added

src/core/others/ojph_file.cpp
```diff
@@ -67,6 +67,13 @@ namespace ojph {
     return fwrite(ptr, 1, size, fh);
   }
 
+  ////////////////////////////////////////////////////////////////////////////
+  size_t j2c_outfile::tell()
+  {
+    assert(fh);
+    return ftell(fh);
+  }
+
```
aous72: Added

src/core/others/ojph_file.cpp
```diff
@@ -191,4 +198,92 @@ namespace ojph {
...
+      //printf("mem_outfile: expanded buffer by %d bytes to %d bytes\n", grow_size, this->size);
+    }
+
+    // copy bytes into buffer and adjust cur_ptr
+    memcpy(this->cur_ptr, ptr, size);
+    cur_ptr += size;
+
+    return size;
+  }
+
```
aous72: Added to code

subprojects/js/CMakeLists.txt
```diff
@@ -14,9 +14,10 @@ set(CMAKE_CXX_FLAGS "-std=c++11 -O3 -DOJPH_DISABLE_INTEL_SIMD")
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/../html)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/../html)
 
-file(GLOB SOURCES "src/ojph_wrapper.cpp" "../../src/core/coding/*.cpp" "../../src/core/others/*.cpp" "../../src/core/codestream/*.cpp" "../../src/core/transform/ojph_colour.cpp" "../../src/core/transform/ojph_transform.cpp")
+file(GLOB SOURCES "src/ojph_wrapper.cpp" "../../src/core/coding/*.cpp" "../../src/core/others/*.cpp" "../../src/core/codestream/*.cpp" "../../src/core/transform/ojph_colour.cpp" "../../src/core/transform/ojph_transform.cpp" "../../src/apps/others/ojph_img_io.cpp")
 add_executable(libopenjph ${SOURCES})
 include_directories(../../src/core/common)
+include_directories(../../src/apps/common)
```
aous72: In my mind, the ojph_img_io.cpp is part of the executable, not the library.

src/core/others/ojph_file.cpp
chafey: whoops, do you want a new PR or did you fix in your merge?  <embarrassed>

subprojects/js/CMakeLists.txt
chafey: Right - I think this was a hack to support some JS hacking I was doing

src/core/others/ojph_file.cpp
aous72: Don't worry.
I removed this function.

subprojects/js/CMakeLists.txt
aous72: That's fine. I will not add it.

src/core/others/ojph_file.cpp
chafey: Wait a sec - this was actually intentional.  It allowed the caller to take ownership of the underling buffer avoiding a copy.  I would much rather use std::shared_ptr<> for this, but since this library is not using the standard c++ library and using c style memory management, this is is necessary.

PR #9, opened 2019-12-25
Do not ignore cmake build type.

boxerab:
Build type can be Release, Debug etc.

Currently, it's not possible to build a Debug version of the
library. Let cmake set the appropriate optimization flags;
for MSVC, just add the /EHsc flag to the existing compiler flags.


aous72:
Thank you Aaron.

PR #14, opened 2019-12-26
Multithreading support

boxerab:
Multithreaded block decoding doesn't improve performance vs single thread. But, there may be performance uplift by multithreading the wavelet transforms. Except, the design may have to change a little for this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,9 @@ cmake_minimum_required(VERSION 3.10.0)
 project (openjph DESCRIPTION "Open source implementation of JPH" LANGUAGES CXX)
 
 set(CMAKE_CXX_STANDARD 11)
+
+find_package(Threads REQUIRED)
+
 if (MSVC)
 	SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHsc /D \"_CRT_SECURE_NO_WARNINGS\"")
 endif()
@@ -10,9 +13,8 @@ endif()
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/../bin)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/../bin)
 
-include_directories(src/core/common)
-include_directories(src/apps/common)
-file(GLOB SOURCES "src/core/coding/*.cpp" "src/core/others/*.cpp" "src/core/codestream/*.cpp" "src/core/transform/ojph_colour.cpp" "src/core/transform/ojph_transform.cpp")
+include_directories(src/core/common src/apps/common)
+file(GLOB SOURCES "src/core/enkitTS/TaskScheduler.cpp" "src/core/coding/*.cpp" "src/core/others/*.cpp" "src/core/codestream/*.cpp" "src/core/transform/ojph_colour.cpp" "src/core/transform/ojph_transform.cpp")
 file(GLOB SSE_SOURCES "src/core/transform/*_sse.cpp")
 file(GLOB SSE2_SOURCES "src/core/transform/*_sse2.cpp")
 file(GLOB AVX_SOURCES "src/core/transform/*_avx.cpp")
@@ -32,8 +34,8 @@ else()
 endif()
 
 add_executable(ojph_expand src/apps/ojph_expand/ojph_expand.cpp src/apps/others/ojph_img_io.cpp)
-target_link_libraries(ojph_expand openjph)
+target_link_libraries(ojph_expand openjph Threads::Threads)
 
 add_executable(ojph_compress src/apps/ojph_compress/ojph_compress.cpp src/apps/others/ojph_img_io.cpp)
-target_link_libraries (ojph_compress openjph)
+target_link_libraries (ojph_compress openjph Threads::Threads)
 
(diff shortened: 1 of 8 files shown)


aous72:
Thank you Aaron for this pull request.

The use of chrono is useful for small files, because it is more accurate, but for 4K, the difference is not that large. 

I have tested multi-threading; it would be a nice addition to the code.  
Currently, most of the time now is spent in the block encoder/decoder; but this only because they are implemented in C/C++.  The performance can improve significantly when they utilize SIMD instructions, and this is possible because the of HTJ2K block coder was designed with SIMD instructions in mind.  However, I do not have the time for it now.  
The improvement of the pull request is modest on 4 core machine in decoding 4K images; I think it would be better to implement faster block encoder/decoder before considering multi-threading.

I think it is useful to leave this as a pull request for anyone who is interested in giving it a try. 

Additional notes:
On Linux, it compiles and works.
On Windows, it did not compile in visual studio 2019 without
```C
#define NOMINMAX
#include <Windows.h> 
```
in "TaskScheduler.h" In fact, it was rather hard to find the problem.
On Windows, the encoder and decoder seem to hang after completion.

boxerab:
> Thank you Aaron for this pull request.
> 
> The use of chrono is useful for small files, because it is more accurate, but for 4K, the difference is not that large.
> 
> I have tested multi-threading; it would be a nice addition to the code.
> Currently, most of the time now is spent in the block encoder/decoder; but this only because they are implemented in C/C++. The performance can improve significantly when they utilize SIMD instructions, and this is possible because the of HTJ2K block coder was designed with SIMD instructions in mind. However, I do not have the time for it now.
> The improvement of the pull request is modest on 4 core machine in decoding 4K images; I think it would be better to implement faster block encoder/decoder before considering multi-threading.
> 
> I think it is useful to leave this as a pull request for anyone who is interested in giving it a try.
> 
> Additional notes:
> On Linux, it compiles and works.
> On Windows, it did not compile in visual studio 2019 without
> 
> ```c
> #define NOMINMAX
> #include <Windows.h> 
> ```
> 
> in "TaskScheduler.h" In fact, it was rather hard to find the problem.
> On Windows, the encoder and decoder seem to hang after completion.

Thanks, Aous. I agree, better to start with SIMD.  
But, I find it strange that there is not a significant speedup by using multiple threads. I will need to investigate further.

boxerab:
So, with some improvements, this branch can now get over 2x performance with threads vs single threaded for decode. To see this consistently, I added a loop to `ojph_expand`; for batch decode, there is a significant speedup. Even so, I am only using around 40% CPU on my 6 core i7 coffee lake, so there is a lot of room for improvement, probably by multithreading the wavelet transform. And, if we add SIMD to the block coder, then we will really be able to see the advance of the new coder over the original Part 1 version.

aous72:
This is good result.  
Now I see what you are trying to achieve -- decoding faster than openjpeg (J2k part 1).
OpenJPH encoder is already faster than openjpeg. The decoder is also faster if openjpeg uses a single thread. This is expected due to how the encoding algorithm are designed.
My ultimate goal is to as fast as turbojpeg, without multithreading.  Kakadu software has already demonstrated this.  But getting there requires developing a 16-bit encoding/decoding path (employing 16-bit instead of the 32-bit/float wavelet transform and block coding currently used).  This is not happening any time soon.

I see the work you have done as useful for anyone interested in a speed up, but I am not very sure how to integrate it. Perhaps, we need to talk further.

Happy new year.

boxerab:
Thanks, Aous, and Happy New Year ! 

Matching turbojpeg is ambitious, but it sounds pretty straightforward to achieve, given the HTJK design, with a combination of multiple threads and SIMD.  If you get this level of performance, then jpeg 2000 could be a true successor to jpeg.  Perhaps some of the turbojpeg developers would be interested in helping out - there is a lot of other work besides performance that needs to be done, such as a unit test framework.

As for integrating mulithreaded decode, I think the code should actually be part of the subband class.
We would need to handle a few different strategies, balancing performance and memory usage. 
To limit memory, only lines of code blocks would be decoded in parallel, or for best performance, all blocks would be decoded in a batch before the wavelet transform.

Jamaika1:
Error: undefined reference to `ojph::local::precinct::alloc_scratch(ojph::local::codestream*)'
```
c:/msys1021/bin/../lib/gcc/x86_64-w64-mingw32/10.2.1/../../../../x86_64-w64-mingw32/bin/ld.exe: ojph_codestream.o:ojph_codestream.cpp:(.text+0xf9fd): undefined reference to `ojph::local::precinct::alloc_scratch(ojph::local::codestream*)'
collect2.exe: error: ld returned 1 exit status
c:/msys1021/bin/../lib/gcc/x86_64-w64-mingw32/10.2.1/../../../../x86_64-w64-mingw32/bin/ld.exe: ojph_codestream.o:ojph_codestream.cpp:(.text+0xf9fd): undefined reference to `ojph::local::precinct::alloc_scratch(ojph::local::codestream*)'
collect2.exe: error: ld returned 1 exit status
```

aous72:
Hi Jamaika1,

I do not intend to support multithreading any time soon.  It may or may not happen.  There are a lot of reasons why.

This code was provided by Arron, and I am thankful for that, but if you needed it, you must fix it yourself, or ask Aaron for his updated version if he has one.

Kind regards,
Aous.

Jamaika1:
Thanks, this is info for someone wasting time. ;)

boxerab:
Closing this - will re-open when project is ready for MT.

PR #17, opened 2019-12-28
precinct: convert scratch to struct member, rather than static

boxerab:
This allows the codestream to be re-used.


boxerab:
Thanks, this issue has been addressed. Closing

PR #27, opened 2020-04-11
build: turn on more warnings, especially sign conversion warnings

boxerab:
Hi Aous,
Hope all is well. In the PR, I've enabled a number of warnings - I think it may be useful to fix these, especially the conversion warnings,  as they often point to bugs in the code.
In my experience, everything should be unsigned unless absolutely necessary :)

Under `gcc`, I saw around 150 warnings, but if you switch compilers to `clang`, you will see a lot more: over 600.

Best,
Aaron


aous72:
Hi Aaron,

Good to hear all is well.  

I agree with you.  Ideally the code should have no warnings.  Warnings can be outright bugs.

I will work my way through them.

Cheers,
Aous.

boxerab:
Great, looking forward to that. 
Aaron

PR #32, opened 2020-05-02
Add cmake submodule support, updates to VSCode settings

chafey:
This PR includes changes to the CMakeLists.txt so it can be used as a submodule by other cmake projects and updates the VSCode settings.  The specific changes to add cmake submodule support are:
1) Making SHARED libraries an OPTION (still defaulting to ON) rather than an explicit setting in add_library.  This lets the consuming CMake project set it to shared or static as needed
2) Exporting the header files via target_include_directories so they can be included directly.  

I am using the git submodule feature to separate the JS/WASM build from the C/C++ build for a number of codecs, you can see how I am doing this for OpenJPH here:

https://github.com/chafey/openjphjs

You may want to consider a similar approach as it provides a cleaner separation of concerns between C++ and JS/WASM, better javascript interface and is going to match the way I am doing this for the other codecs such as CharLS, OpenJPEG and TurboJPEG.


aous72:
Hi Chris,

This [page](https://chafey.github.io/openjphjs/test/browser/index.html) is really impressive.  I am so excited that my code is useful.  I might link to it, or maybe host it somewhere with links to your repository.  What do you think?

The pull request looks alright to me.  I will integrate it into version 0.7.0, for which the main focus is to remove any warnings in Linux and Windows.

I also think having emscripten bindings is perhaps a better route.

Thank you Chris.

Kind regards,
Aous.

chafey:
Thank you!  You can certainly link to it and of course fork/copy/whatever (its open source!).  I want to do a lot more demos like this including one that compares performance between multiple codecs (HTJ2K, JPEG2000, JPEGLS, JPEG Baseline, RLE, gzip, etc) and one that does interactive encoding where you can adjust various parameters and see how it affects image quality and size in real time.

aous72:
Hi Chris,

I had a look at the code.  You are very familiar with the code.

I have a slightly different idea.  I will add this functionality this week.

The idea that I have is to add a 
"codestream.restrict_resolution(int num_res_to_skip_for_data, int num_res_to_skip_for_recon)" 
function call specifying the number of resolutions to skip. 
num_res_to_skip_for_data specifies the number of resolutions for which data reading data is skipped.  During data parsing, precincts that belong to skipped resolutions are not read.
num_res_to_skip_for_recon which should be smaller than num_res_to_skip_for_data, specifies the resolution at which image is reconstructed.  This is useful for simulating progressive decoding similar to JPEG.  I think you mentioned this to me before; it took me a bit of time to absorb it.
This way all progression orders should be supported.

I might do one version before then, because I have version 0.7.0 in development.

Thank you Chris for the encouragement.

Kind regards,
Aous.

chafey:
You have done excellent work, I am glad that it is encouraging!  Will your approach work with an incomplete bitstream? I updated the web page to support sub resolution decoding to help understand how progressive loading would work:

https://chafey.github.io/openjphjs/test/browser/index.html

aous72:
It is definitely on my mind.

I think I need to add of flag to make the decoding more resilient.  If there is something that cannot be decoded, it should be skipped to decoding the next decodable part, if that exist.  In the end, the decoder will generate something useful, even if there is an error in the codestream.

This way we should be able to decode incomplete codestreams.

chafey:
Yikes - just now realized that my commits to master unrelated to this PR show up here.  Sorry about that, I should have used a branch!  Do you want me to create a new PR on a branch to facilitate merging or can you add the CMake changes manually?

aous72:
Sorry, I did not get that.

In the earlier pull request, commit c857be9, you have changes to 3 files, CMakeLists.txt and two related to VS code.  I have integrated them into development.  I think that CMakelists.txt was fine.  The ones related to VS Code, I am do not know.  Are these the ones you are talking about?

chafey:
Yes - those are the ones!  I'll close this PR, thanks!

PR #38, opened 2020-05-27
ojph_arch: no need to XOR with 31

boxerab:
Hi Aous,
Just a small fix for windows - _BitScanForward calculates trailing zeros directly.
Regards,
Aaron


aous72:
Thank you for adding this.

I never used the function, so I never noticed the issue.

boxerab:
> Thank you for adding this.
> 
> I never used the function, so I never noticed the issue.

My pleasure. I think you need a test suite :) 
Thanks for merging.

PR #39, opened 2020-08-31
Added CMake commands for development.

normanb:
This PR adds library versioning on Linux and adds the installation steps required to install a set of development header files in a standard location.


aous72:
Hello Norman, 

Thank you for the effort and taking the time to put this pull request in.

Pardon my ignorance, I am not very familiar with cmake.  However, I know what you are trying to do.

I am happy with the pull request, but I have a couple of questions.
Do you object to having the executables/library in ../bin? if not, I might but this back in.
I had an error in line 102, regarding the ARCHIVE, which I will fix after this pull request by replacing ARCHIVE with Library.
We only need the header files in core/common; I will make the necessary changes.
I will also include ojph_version.h in my ojph_def.h header.

Cheers,
Aous.

normanb:
Hi Aous,

Thanks for reviewing the PR, I moved the executables/library to the build folder so that I could do `rm -rf *` safely within build on the command line, but it is minor and ../bin is fine.

Thanks for merging!

PR #41, opened 2020-10-09
ojph_decode_codeblock: return false on error condition

boxerab:
Hi Aous,
I hope all is well ! Here is an issue I found today.

Simply returning without error leaves the code block
in an undefined state. Better to let the caller know that
something went wrong.

Best Regards,
Aaron


aous72:
Hi Aaron,

I hope you are doing well.

Thank you for putting this in. I agree with you.

Kind regards,
Aous.

boxerab:
Thanks, Aous.

PR #47, opened 2020-11-22
Add a few sanity checks on decoder to avoid illegal left shift

boxerab:
Hi Aous,
Hope all is well. Here are a few changes I made to the block decoder to avoid an illegal left shift
I am seeing on some corrupt code streams. What do you think?
Thanks,
Aaron

### DIFF
--- a/src/core/codestream/ojph_codestream.cpp
+++ b/src/core/codestream/ojph_codestream.cpp
@@ -38,6 +38,7 @@
 
 #include <climits>
 #include <cmath>
+#include <stdexcept>
 
 #include "ojph_file.h"
 #include "ojph_mem.h"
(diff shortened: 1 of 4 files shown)


aous72:
Hi Aaron,

Thank you for putting this in.  I was busy with other work today.  I will get back to you tomorrow.

I need to think about some of these issues, and I may have a couple of questions.  I agree on the left shift issue, and I think we can have a more strict condition.

Kind regards,
Aous

boxerab:
Thank you , Aous. Looking forward to your feedback.
Aaron

aous72:
Hi Aaron,

These are my thoughts.

#### For "decode: fail on illegal left shift"
I propose to check U_q once they are evaluated at line 1171
```
ui32 consumed_bits = decode_init_uvlc(vlc_val, uvlc_mode, U_q);
```
and line  1419
```
ui32 consumed_bits = decode_noninit_uvlc(vlc_val, uvlc_mode, U_q);
```
using 
```
if (U_q[0] > missing_msbs || U_q[1] > missing_msbs)
  return false;
```
I think return false is fine.  Is there a reason why prefer to through an exception?
The return value should be tested and if the return value is false and the decoder is running in "resilient" mode, the block should be filled with zeros and decoding continued, otherwise the code should produce an error message and terminate.

Another worthwhile test is for p, line 1014
```
ui32 p = 30 - missing_msbs; // The least significant bitplane for CUP
```
p should be 1 or more.  If p is 1 then the number of passes should be forced to 1 (32 bits are not enough for more passes).
It is conceivable to handle the case of p = 0, but this require a new path in the code.

#### For " MEL decode: throw exception in addition to assert for added code stability"
The assert ```assert(melp->unstuff == false || melp->data[0] <= 0x8F)``` fails when the bitstream does not comply with the stuffing rules, and therefore it is malformed.  This does not crash the program, but may produce rubbish image.  Checking for such errors in the codestream is possible but comes with an overhead.  My thought is not to test for them.  It is conceivable to write a different code that checks all these malformed situations, but such code runs slower.  I think we should only test for cases that may crash the program.  What do you think?

Kind regards,
Aous.

aous72:
I push the proposed modifications.  Let me know what you think.

boxerab:
Thanks, Aous - this is a much better fix. It also solves another heap corruption that I was seeing. I will close this.

PR #51, opened 2021-01-01
fix: writing 2 bytes into a region of size 1

Jamaika1:


aous72:
Hi Jamaika1,

First, happy new year.

Sorry I did not come back to you earlier. 
But in the earlier issue (issue #50, which is now closed), the corrections you made did not make sense for me. TPsot andTNsot are 8 bits fields and I am writing one byte to them.

In this issue, the program's idea is that if an error happens, all fields are set to zero, instead of leaving them in indeterminate states.  Having types before them defeats this purpose.  So I am not merging this.

I suspect, this "writing 2 bytes into a region of size 1" is a bug in the compiler.  Could you please let me know what compiler are you using?

Could you please also write some text to explain what you are thinking? This makes it easier for me.  Thank you.

Feel free to discuss these issues with me.

Kind regards,
Aous.

Jamaika1:
Happy New Year

Let me say it this way. Out of New Year's Eve boredom, I tried to figure out why the quantization step size function was not working.
How to set the codec to lossless under gcc 10.2.
It's true that gcc doesn't like undefined variables. GCC has a lot of fixes however and bugs since. The new year seems like a year of big codec changes and news, but it can be a year of apathy. I can't compile properly because there's nowhere to get the new gcc from.
http://msystem.waw.pl/x265/

cht_j2c.exe -i image_21447_24bit.yuv -o image_21447_24bit.j2c -dims {1563,1558} -num_comps 3 -signed false -bit_depth 8 -downsamp {1,1},{1,1},{1,1} -block_size {64,64} -precincts {128,128},{256,256} -prog_order CPRL -qstep 10.0
_Assertion failed: coded_cb->missing_msbs > 0, file ojph_codestream.cpp, line 3933_

I don't count too much in the new year and the income is what it is. However I am counting on some updates.

Kind regards,
Lucas

aous72:
Hi Lucas,

For lossless coding, OpenJPH does not support quantization yet; you can only get the full quality image.

For lossy coding, quantization is supported and I believe it is working.  -qstep cannot be 10.0 however.

We think of the input image as being in the range [-0.5, 0.5).  That is, the input data x, is such that -0.5 <= x < 0.5.  If your data is unsigned 8-bit data, then the code will converted it to signed  8-bit in the range [-128, 127], and then divide it by 128, which is 2^(8-1).  For 12-bit, the same is true, subtract 2048, and divide by 2048.  Signed data is kept as signed data; no subtraction is needed.

qstep >= 0.5, will render all coefficient to 0 or 1 ( at 0.5 some coefficient might not be zero because of BIBO gain). If you do not specify qstep, then the code uses 
```
1.0f / (1 << (siz.get_bit_depth(0) + siz.is_signed(0)))
```
Note: I think I should remove the addition of is_signed, but the code still works.  

For unsigned 8-bit the default qstep value is 0.0039.  So, to have smaller files, try 0.01, 0.05, 0.1, 0.2.  To have larger files you can go to 0.0001 for example.

Feel free to contact me if you have more issues.

I know I have been very slow here.  
There is another task I am working on, which is more important to me, but it has been a slow grind.

Cheers,
Aous.

Jamaika1:
Thanks for the answer
Maybe at least change the description in help.
_-qstep        (1.0/255,0) quantization step size; all quantization
               step sizes are derived from this._
Why is the quantization step size one by default?

aous72:
Yeah, I admit it is confusing and I should change it.  What I meant is that 
qstep = 1/255 = 0.0039.

I will change this in my development branch.

Cheers,
Aous.

Jamaika1:
What does the minus value mean?

aous72:
what do you mean? qstep has to be positive.

Jamaika1:
Default is -1.0

aous72:
This is for me to detect if it has been supplied by the user.  
In param_qcd::check_validity() I see it to an appropriate value if it is still -1.0.

Jamaika1:
Now I understand. Now you're waiting for lossless.

aous72:
I do not get you.  For lossless, you have to use 
-reversible=true

Jamaika1:
Extra. It works.

PR #52, opened 2021-01-02
Change of description

Jamaika1:


src/apps/ojph_compress/ojph_compress.cpp
```diff
@@ -461,9 +461,9 @@ int main(int argc, char * argv[]) {
 
     "The following option has a default value (optional):\n"
     " -num_decomps  (5) number of decompositions\n"
-    " -qstep        (1.0/255,0) quantization step size; all quantization\n"
-    "               step sizes are derived from this.\n"
-    " -reversible   (false) for irreversible adnd true for reversible\n"
+    " -qstep        (-1.0...1.0) quantization step size; all quantization without 0.0\n"
```
aous72, line 464: -qstep      (0.0039 for 8-bit data) quantization step size; quantization step sizes for all subbands are derived from this value. qstep value should be from a very small positive float number such as 0.00001 and up to 0.5.  Larger values produce smaller files (i.e., coarsely quantization).

I think there is no need to mention the -1.0; it is just an implementation issue that I might change,

src/apps/ojph_compress/ojph_compress.cpp
```diff
@@ -461,9 +461,9 @@ int main(int argc, char * argv[]) {
...
+    "               step sizes are derived from this. {default -1.0}\n"
+    " -reversible   (false) for irreversible lossy and true for reversible lossless\n"
```
aous72, line 467: -reversible   (false) for irreversible; that is, lossy compression using the 9/7 wavelet transform; and true for reversible compression; that is, lossless compression using the 5/3 wavelet transform.

src/apps/ojph_compress/ojph_compress.cpp
```diff
@@ -461,9 +461,10 @@ int main(int argc, char * argv[]) {
 
     "The following option has a default value (optional):\n"
     " -num_decomps  (5) number of decompositions\n"
-    " -qstep        (1.0/255,0) quantization step size; all quantization\n"
-    "               step sizes are derived from this.\n"
-    " -reversible   (false) for irreversible adnd true for reversible\n"
+    " -qstep        (-1.0000...1.0000) quantization step size for lossy\n"
```
aous72, line 464: -qstep  (0.0039 for 8-bit data) quantization step size; quantization step sizes for all subbands are derived from this value.  qstep values can be between a very small positive float number such as 0.00001 and 0.5.  Larger values produce smaller files (i.e., coarsely quantized coefficients).

I think there is no need to mention the -1.0; it is just an implementation detail that I might change.

Jamaika1: Too bad there is no printf. Could use ± sign

Jamaika1: I made changes as I understood.

aous72:
That's great.  

Thank you for pointing out this issue.

Cheers,
Aous.

PR #53, opened 2021-01-11
Change description

Jamaika1:


src/apps/ojph_compress/ojph_compress.cpp
```diff
@@ -468,11 +468,9 @@ int main(int argc, char * argv[]) {
     "               lossy compression using the 9/7 wavelet transform;\n" 
     "               and true for reversible compression; that is,\n"
     "               lossless compression using the 5/3 wavelet transform.\n"
-    " -colour_trans (true) if there are three color components that are\n"
+    " -colour_trans (-1) if there are three color components that are\n"
```
aous72, line 471: Hi Lucas,

Thank you for your continued check on the code.

-1 is for me to detect if the user has specified something.  The actual value is either true or false.  The default is true if you know that you have three color components that are known to be RGB.
The color transform converts RGB to YUV as specified in ITU-R BT.601; therefore, this transform does not make sense when the original file is already in YUV.  This transform improves coding efficiency.  

Perhaps, one can add that "This option is not available for yuv files."

By the way, the library should support rgba, but ojph_compress does not read these files yet.

Cheers,
Aous.

aous72: Hi Lucas,

Sorry if I was confusing.  This is my thinking.

true if you want color transform, false if you do not want.

Color transform converts rgb to yuv.  Then, when image is yuv, we cannot employ color transform.

> ojph error 0x01000031 at ojph_compress.cpp:684: we currently do not support color transform on yuv files. In any case, this not a normal usage scenario. The OpenJPH library however does support that, but ojph_compress.cpp must be modified to send all lines from one component before moving to the next component; this requires buffering components outside of the OpenJPH library

I agree that this is confusing.  I think I should remove _The OpenJPH library however does support that, but ojph_compress.cpp must be modified to send all lines from one component before moving to the next component; this requires buffering components outside of the OpenJPH library_.  

Currently, I assume yuv is 1 or 3 color components.  We can easily (not tested) modify this to make it 4 components.  The Library itself should support a large number of components, but I never tested with more than 3.  Therefore, I may have bugs.  

I also assume that .yuv are in YUV or YCbCr color space, and therefore cannot employ color transform, because original is not in RGB.

I think I should have raw input files. In this case, all possible raw inputs can be handled including more than 3 components. Perhaps, I should leave it to the user to decide if (s)he wants color transform; not force anything on the user.

Let me know your thoughts.

Cheers,
Aous.

Jamaika1: _I also assume that .yuv are in YUV or YCbCr color space, and therefore cannot employ color transform, because original is not in RGB._
The joke right now is that no photo viewer I know can handle jpeg2000-yuv. Five years ago it was understandable because the world was 8bit and nobody wanted innovation. Now what to do with heif/avif 10<->12bit hdr images? Not popular, but it's for Windows. I read that jpeg 2000 part 16 is compatible with heif.

ValZapod: > No photo viewer I know can handle jpeg2000-yuv444p

ffplay, mvp that is based on ffmpeg. Enough for me. ;)

ValZapod: Just checked, Photoshop 2021 also supports it. Though it does not say it is YUV. Ha.

Also very funny is that exiftool thinks it is sYCC (it is not YUV and is not sRGB, it is extended gamut sRGB in YUV, where out-of-gamut, per sé negative RGB values are further defined with XYZ to linear RGB matrix).

Also, double checked Chrome does not support it too. Sigh.

aous72:
That is an interesting conversation.

> Does jpeg2000 have a color sampling header?

I am not super sure, but my understanding is as follows:
There is the jpeg2000 codestream; this codestream has the extension .j2c
The codestream is wrapped with another header(s).  The complete file (codestream + wrapper) has the extension .jp2 for JPEG2000 part 1 of the standard (this the jpeg2000 commonly available), .jpx for JPEG2000 part 2 (has extended features such as user defined wavelets, a different wavelet decomposition structure, etc.), .mj2 for video (each image coded on its own, but all are put together in one file), and .jph for HTJ2K, also known as Part 15 of the standard; my code generates the codestream for .jph, but not the wrapper header.

These wrappers (you can think of them as headers) can contain information about the color space of the image, and other information.  A proper decoder should be able to use this information to tell how to correctly generate the result.  The .j2c I generate does not have these wrappers, so you can not be very sure what to do with color components in the codestream, but what I produce is more like a learned guess.
The codestream has a bit to say perform color transform.

The codestream itself has information about how each component in the file is subsampled (inside a header marker called SIZ); the codestream also has another marked called CRG (component registeration); this field tells the decoder the amount of shift that should be employed for each component.  This is useful to distinguish the two cases shown below, taken from [here](https://www.pcmag.com/encyclopedia/term/chroma-subsampling)
chroma-subsampling-ycbcr420 fit_lim size_800x
To reconstruct full resolution image, different interpolation filters are needed.

I hope this answers your question; please as it again in a different way.

Cheers,
Aous.

PS If you are interested in JPEG2000, have a look at kakadusoftware at
https://kakadusoftware.com/downloads/
They share binary tools of their library for Windows, Mac, and Linux.  GUI for media viewing is available for Windows and Mac, but not Linux.  The shared tools support htj2k (or .jph), but the shared version do not have the really fast code for .jph.

ValZapod:
BTW, will attach such a file here (the colors are a little wrong because the file was supposed to be in BT.601 525 primaries and with encoding gAMA png chunk 1.8007, but i did not bother to transfer them).

[relaxout.zip](https://github.com/aous72/OpenJPH/files/5863051/relaxout.zip)

aous72:
Hi Lucas,

I am merging this with the development branch before modifying it.  I will try to keep the main message you are trying to say.

Cheers,
Aous.

PR #56, opened 2021-01-20
FEAT: Subresolution decoding

chafey:
This PR adds support for sub resolution decoding.


chafey:
My mistake - this is old code and desired functionality is already available via restrict_input_resolution()

PR #60, opened 2021-02-06
Truncated stream

boxerab:
Hi Aous,
Here is a small change that allows the decoder to handle a truncated file. I probably didn't handle all of the cases
where the file IO hits EOF - let me know if you think this change makes sense for the library.
Kind Regards,
Aaron


aous72:
Hi Aaron,

Thank you for the effort.

I have a "resilient" mode, where the decoder attempts to decode all the data it has, even if the codestream is truncated too early.
This can be set using ```void codestream::enable_resilience()```.  If I understand things correctly, this commit is in this direction.

Let me know what you think.

Kind regards,
Aous.

PR #73, opened 2021-09-24
Fix wrong error mesg when a netpbm files with P6 keyword has no a .ppm extension

rouault:


aous72:
Thank you Even.

aous72:
Too much copy and paste.

On another note.  I have a development branch, which has a few bug fixes, and OpenJPEG work is based on that branch.  If you want to fiddle with the code, I suggest you use that.

PR #74, opened 2021-09-25
Fix enable TIFF

Jamaika1:


src/apps/ojph_compress/ojph_compress.cpp
```diff
@@ -644,7 +644,7 @@ int main(int argc, char * argv[]) {
         base = &ppm;
       }
       #ifdef OJPH_ENABLE_TIFF_SUPPORT
-      else if ((strncmp(".tif", v, 4) == 0) || (strncmp("tiff", v, 4) == 0))
+      else if ((strncmp(".tif", v, 4) == 0) || (strncmp(".tiff", v, 4) == 0))
```
aous72, line 647: Thank you for spotting this error. Could you please change 
```
(strncmp(".tiff", v, 4) == 0))
```
to
```
(strncmp(".tiff", v, 5) == 0))
```
That is compare 5 characters instead of 4.

PR #75, opened 2021-09-25
Fix enable raw & tiff

Jamaika1:


aous72:
Same as before merge and modify.

PR #77, opened 2021-10-05
FIX: CMAKE_CXX_FLAGS caused compile failures for Clag|GNU

chafey:


aous72:
Hi Chris,

Thank you for putting this in, and for your continuous support.

Kind regards,
Aous.

PR #80, opened 2022-03-30
Prefer use of cmake module GNUInstallDirs to install libs

malaterre:


malaterre:
@aous72 I've submitted a series of patch we use for openjph package in Debian:

* https://salsa.debian.org/debian-phototools-team/openjph/-/tree/master/debian/patches

Please review/comments individual PR, thanks !

PR #81, opened 2022-03-30
Fix conflict with naming convention of shared library

malaterre:


malaterre:
@aous72 I guess you do not want to pull this PR, but this is to get the discussion started.

aous72:
So how to fix this, or what's wrong with it?

malaterre:
> So how to fix this, or what's wrong with it?

This breaks the usual mechanism of SONAME on linux. Do you really:

1.  mean to encode the full version in the library filename,
2. or are you ok to remove portion of it ?

Both solutions are acceptable.

malaterre:
Currently without this patch, I get:

```
$ DESTDIR=/tmp/i make install
[...]
-- Installing: /tmp/i/usr/local/lib/libopenjph.0.8.2.so.0.8.2
-- Installing: /tmp/i/usr/local/lib/libopenjph.0.8.2.so.0.8
-- Installing: /tmp/i/usr/local/lib/libopenjph.0.8.2.so
```

where:

```
 % readelf -d /tmp/i/usr/local/lib/libopenjph.0.8.2.so.0.8.2 | grep soname
 0x000000000000000e (SONAME)             Library soname: [libopenjph.0.8.2.so.0.8]
```

Setting of the full version *in* the filename defeat the whole mechanism. I think you can remove the PATCH version from the filename.

malaterre:
@aous72 I've implemented option (4) above.

aous72:
@malaterre 
Thank you for the effort and putting this in.  I agree with you having repeated 0.8.2  in the name does not make sense.

After testing with windows with this PR, all versioning numbers in the file are lost. 
I think the version section should be:
```
if (OPENJPH_VERSION AND UNIX)
  set_target_properties(openjph
    PROPERTIES
      SOVERSION "${OPENJPH_VERSION_MAJOR}.${OPENJPH_VERSION_MINOR}"
      VERSION "${OPENJPH_VERSION}")
#else()
  set_target_properties(openjph
    PROPERTIES
      OUTPUT_NAME "openjph.${OPENJPH_VERSION_MAJOR}.${OPENJPH_VERSION_MINOR}")
endif()
```
Is that alright with you?

Final question:
With this, do I needed to preserve Application Binary Interface "ABI" when "OPENJPH_VERSION_PATCH" changes?

Thank you.

PR #85, opened 2022-04-08
Fixed bug where libopenjph_simd was linking to non simd library

chafey:


aous72:
@chafey Thank you for the bug fix.

PR #86, opened 2022-05-22
Added OJPH_STATIC

Jamaika1:


src/core/common/ojph_arch.h
```diff
@@ -200,7 +200,7 @@ namespace ojph {
   /////////////////////////////////////////////////////////////////////////////
   // defines for dll
   /////////////////////////////////////////////////////////////////////////////
-#ifdef OJPH_OS_WINDOWS
+#if defined(OJPH_OS_WINDOWS) && !defined(OJPH_STATIC)
```
aous72, line 203: Hi Lucas,

Thank you for this.

Change this to
```
#if defined(OJPH_OS_WINDOWS) && defined(OJPH_BUILD_SHARED_LIBRARY)
```

You also need to change top level CMakeLists.txt to include 
```
if (BUILD_SHARED_LIBS AND MSVC)
  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /D \"OJPH_BUILD_SHARED_LIBRARY\"")
endif()
```

I have a big merge coming.  I will delay this.

Cheers,
Aous.

PR #87, opened 2022-06-23
Options to insert a TLM marker, and create tileparts

aous72:
Added options to enable:
1. the insertion of a TLM marker. 
2. the creation tileparts at resolution, component, or resolution and component levels.  This feature is not allowed to occur at precinct boundaries if they occur before the needed component or resolution levels. Depending on the progression order, this means that this feature is limited or missing; for example, inserting tileparts in the PCRL progression means each precinct would have its own tilepart.  While this is possible, one has to keep in mind that we cannot have more then 255 tileparts in a tile; therefore, no tileparts in this progression order is allowed.

Please leave a comment on what you think.


palemieux:
Seems to be working. See sample J2C and basic TLM parser:
[ojph.zip](https://github.com/aous72/OpenJPH/files/8971884/ojph.zip)

`./bin/ojph_compress -i 00017.tif -o ojph.j2c  -reversible true -prog_order RPCL -block_size \{64,64\} -precincts \{128,128\},\{256,256\} -num_decomps 5 -tlm_marker true -tileparts R`

aous72:
Thank you for independently confirming that it works correctly.  The script is useful.

PR #104, opened 2023-05-08
Walk around overflow when block width is bigger than 64

RodneyGuo:
This is a quick around to the overflow issue.
This WA may cause performance issue
due to high useage of the ZMM register.

### DIFF
--- a/src/core/coding/ojph_block_encoder_avx512.cpp
+++ b/src/core/coding/ojph_block_encoder_avx512.cpp
@@ -1035,26 +1035,6 @@ void ojph_encode_codeblock_avx512(ui32* buf, ui32 missing_msbs,
     //For a 1024 pixels, we need 512 bytes, the 2 extra,
     // one for the non-existing earlier quad, and one for beyond the
     // the end
-    /* The maximum size of the width is 64 which is defined in the 
-     *  ojph_params_local.h
-     *
-     * struct param_cod {
-     *     SPcod.block_width = 4;
-     *
-     *     size get_block_dims() const
-     *     {
-     *         return size(1 << (SPcod.block_width + 2),
-     *                     1 << (SPcod.block_height + 2));
-     *     }
-     * };
-     *
-     * So we use 3 * ZMM registers to save the e_val
-     * The 3rd ZMM is used to save the shifted bytes.
-     *
-     * If the maximum size of the width is changed in the future.
-     * Do remember to change this implementation.
-     *
-     */
     const __m512i right_shift = _mm512_set_epi32(
       0, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1
     );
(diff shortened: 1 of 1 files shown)


aous72:
Thank you.  Let's merge it and I can give it a try.

It looks like register allocation is a difficult problem.
In general, very wide codeblocks are possible, but not very common.

aous72:
Thank you.  It is passing all tests.
I will merge it with master.

aous72:
A crude, yet simple way of sorting this out is to have different versions for different widths.

We create a template whose only parameter is the number of elements in e_val_vec, using 2 for a width of 32 of smaller, 3 for width of 64, 5 for 128, 9 for 256, 17 for 512, and 33 for 1024.

This enables excellent performance for block widths that are not large, and lower performance for wider blocks, which are not very common.  Basically paying more attention to the more important case.

Cheers,
Aous.

RodneyGuo:
Thanks Aous,

I did a quick experiment with our local test cases. Seems this walkaround did not cause performance impact for small block width cases.
I will try to optimize it by saving data by bytes instead of by dword in ZMM register. The number of the ZMM register used will be decreased but addition instruction are needed. Not sure if it will have positive effect or negative effect.

BTW:
Is it possible the block width is bigger than 1024? I saw in the non-AVX implement, a buffer of 512 is given which an support maximum 1024 block width.

aous72:
1024 is the widest.
The largest block has 4096 samples, which can be 1024x4 or 4x1024. All other sizes, such as 16x256, are possible so long that they do not exceed 4096 samples.  4 is also the smallest width or height, but it can be restricted further, using some other settings, to 1 or 2 samples.

RodneyGuo:
> 

This is really good idea.

aous72:
Overnight I realized that it might be useful to discuss what ```__m512i e_val_vec[33];``` means.
It is a trap that took some time for me to get my head around during my GPU programming.

This declaration is store in RAM, not in registers, unless certain conditions are met.
Because it is stored in RAM, you see the segmentation fault and the no degradation in performance when you changed [3] to [33].
It is stored in RAM, because registers are not index-able.  That is, if ```e_val_vec[0]``` is stored in zmm0, what does ```e_val_vec[i]``` mean? ```e_val_vec[i]``` cannot mean ```zmm0 + i``` because the hardware does not support that.  Therefore, the only way to achieve indexing is by using RAM.

For it to be stored in registers, either of two conditions must be met.
The first, you always access ```e_val_vec``` with constant indices; in this case ```e_val_vec[0]``` means ```zmm5``` and ```e_val_vec[10]``` means ```zmm4```, and so on.  The declaration ```__m512i e_val_vec[33];``` becomes a convenient way of declaring ```__m512i e_val_vec0, e_val_vec1, e_val_vec2``` and so on.
The other possibility is if the compiler unrolls the loops creating an implementation where the index is fixed and known.

Hope this helps.

Cheers,
Aous

aous72:
I agree with your statement. I only have two comments.
1. zmm are 32 registers; there are also mask registers k0-k7.  
2. I agree that the declaration __m512i e_val_vec[33]; does not specify the location, but depending on usage, the compiler might be forced to store it in RAM/cache.  This was discussed by my earlier comment https://github.com/aous72/OpenJPH/pull/104#issuecomment-1539252186

PR #105, opened 2023-05-23
Add support for writing a user-specified COM marker

palemieux:


aous72:
Thank you for this pull request.

A couple of questions comes to my mind and I need your opinion.
1. What if you want more than one comment.  One possibility is to pass "char *argv[]", with argc, like approach. Or is that very unlikely.
2. I usually try to limit the number of new/delete within the library; I mostly employ the two memory allocators elastic and fixed. I use new in only a few other places within the library, usually on unlikely paths -- COC was a hack, and I think I figured out how to not use "new" there.  I care less about the supporting code, and I do not worry as much using new/delete there; supporting code such ojph_compress and ojph_expand, img_io, other structures external to the library.
To this.  One can rely on the external application keeping the string memory buffer alive for the duration of using the library.  This case the library do not need to create an internal buffer; your solution is neater, but this solution should be faster.  What do you think?

If we use 1 and 2, we can have multiple comments and we do not need to allocate memory.  ojph_compress can use allocate/deallocate memory to accommodate a number of comments, perhaps starting with 3 or 4 and expanding that if it find more.

Thank you.

You do not have to change the code.  I can merge it with development and then edit it.

palemieux:
> What if you want more than one comment. One possibility is to pass "char *argv[]", with argc, like approach. Or is that very unlikely.

It would be nice, and can propose code, but no hard requirement from me at this point.

> One can rely on the external application keeping the string memory buffer alive for the duration of using the library.
> [...]
> What do you think?

What if the COM marker changes every codestream and there are 200,000 codestreams created? The caller could be responsible for keeping the memory alive for one codestream, but probably not for the lifetime of the library. Does the caller control when it is done with a codestream?

Does one "new" per codestream really impact speed?

aous72:
> What if the COM marker changes every codestream and there are 200,000 codestreams created? The caller could be responsible for keeping the memory alive for one codestream, but probably not for the lifetime of the library. Does the caller control when it is done with a codestream?

Maybe I misspoke. I meant the need to be alive until the codestream is generated.
I thought about your question of caller control.  See below.

>Does one "new" per codestream really impact speed?

No, but my worry is "new creep" when we keep adding them.  An alternative solution is to use the elastic memory allocator for this.

I looked at the code again, and realized that currently the comment is written when the caller calls ```write_headers```. So, we can potentially make it part of that call.  This allows us to add a number of comments, as explained earlier, if desired.
I also looked at the standard, it seems that comments can be added to the main header, but also to every tile part; it looks like even Kakadu does not support adding comments to tile parts.  It is then fine if we do not support adding comments to tile parts.

I will think about it.  Either elastic, creating a linked list of comments, or in the write_headers call.  What are your thoughts?

palemieux:
> in the write_headers call

+1

aous72:
Ok then. I will do it tonight.

aous72:
Perhaps use (int num_comments, char* comments[]).

palemieux:
it gets messier (and may require API changes) if the `Rcom` parameter changes depending on the COM marker.

aous72:
Thank you for bringing this up. I did not notice it earlier.
For a good implementation, we need to specify the type of data it is (0, 1) and the length because we cannot rely on the null-terminated string.  It possible to create a structure and ask the caller to fill-in these structures before calling write_hearders.

Perhaps, we do a single null-terminated string now, and we leave the more complicated option, which requires more fields, to later, if somebody wanted it.  This, of course, would require a different function call. What do you think?

michaeldsmith:
I've found both binary and text com values useful. In past applications, I've used binary comments to store the entire header of the original file in a lossless compression workflow in combination with lossless J2K compression. This allows a special decoder that is looking for the header info in a COM marker to reconstruct the original file exactly, and a regular decoder that ignores the COM marker will just decode the pixel values. 

-----Original Message-----

aous72:
Thank you Mike.

How about passing this structure to write_headers
```
  class comment_exchange
  {
  public:
    param_com() : data(NULL), len(0), Rcom(0) {}

    void set_string(char *str)
    { data = str; len = strlen(str); Rcom = 1; }
    void set_data(char* data, ui16 len)
    { this->data = data; this->len = len; Rcom = 0; }
    
  private:
    char *data;
    ui16 len;
    ui16 Rcom;
  };
```
to 
```
void codestream::write_headers(outfile_base *file, comment_exchange* comments, int num_comments)
```
For Pierre's application, you only need to pass argv[i] to set_string().  The comment_exchange object can be allocated on the heap.

palemieux:
`comment_exchange* comments` -> `const comment_exchange* comments`

PR #117, opened 2024-01-10
trivial typo fixes

bradh:


src/apps/ojph_compress/ojph_compress.cpp
```diff
@@ -900,7 +900,7 @@ int main(int argc, char * argv[]) {
         cod.set_progression_order(prog_order);
         if (employ_color_transform != -1)
           OJPH_ERROR(0x01000086,
-            "color transform is not meaningless since .raw files are single "
+            "color transform is meaningless since .raw files are single "
```
bradh, line 903: Or "...is not meaningful..." which is harder to understand but maybe softer language.

aous72: Hi Brad,
Thank you for catching all these typos.
Kind regards,
Aous.

I am fine with meaningless here.

PR #118, opened 2024-01-11
add getter for TLM marker request

bradh:
Small code addition to check whether TLM marker has been requested.

Without this, I need to have a shadow variable to check whether this property is set.


aous72:
All good. Thank you.

PR #119, opened 2024-01-11
make comment_exchange string const

bradh:
The comment (string/data) is not modified, and can be `const`. This makes it easier to pass in something from C++ std::string.

This would be an API change, so OK if you'd prefer to drop it.


aous72:
All good. Thank you.

PR #120, opened 2024-01-12
add accessors for tilepart division

bradh:
This exposes the current tilepart division setting. This avoids needing a shadow variable in the libheif plugin.

Also fixes a couple of typos in the applicable enumeration. I can split that out as a separate PR if preferred.

### DIFF
--- a/src/core/codestream/ojph_codestream.cpp
+++ b/src/core/codestream/ojph_codestream.cpp
@@ -109,6 +109,20 @@ namespace ojph {
   }
 
   ////////////////////////////////////////////////////////////////////////////
+  bool codestream::is_tilepart_division_at_resolutions()
+  {
+    ui32 res = state->get_tilepart_div() & OJPH_TILEPART_RESOLUTIONS;
+    return res ? true : false;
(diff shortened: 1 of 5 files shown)


aous72:
Hi Brad,

This is all reasonable.

However, it is worth saying that when a tile bitstream is partitioned into multiple tile-parts, the only requirement is that each tile part contains complete precinct bitstreams; that is, a precinct bitstream cannot be broken across two tile parts.

While OpenJPH can create tile parts in a couple of specific ways (resolution, component), as you have noticed, other tools may partition the bitstream in other ways.  Would that be a problem for libheif?

Cheers,
Aous.

bradh:
> While OpenJPH can create tile parts in a couple of specific ways (resolution, component), as you have noticed, other tools may partition the bitstream in other ways. Would that be a problem for libheif?

This is only on the libheif encoder side (i.e. setting it as an option parameter, and reading that parameter back), so I am confident it won't be a problem at this stage.

Decoding is handled separately - its a whole different plugin.

PR #123, opened 2024-01-27
include ojph_arch.h in ojph_codestream.h and ojph_params.h to compensate for OJPH_EXPORT definition

bradh:
These headers use `OJPH_EXPORT` but do not include the `ojph_arch.h` header that provides it. That potentially causes compile-time errors when using those headers, depending on what other headers are used and the order of inclusion.


aous72:
Hi Brad,
Thank you for your contribution.

PR #145, opened 2024-07-02
Fix includedir/libdir in pkg-config file when CMAKE_INSTALL_INCLUDEDIR/CMAKE_INSTALL_LIBDIR are absolute paths

abustany:
This fixes the pkg-config file generation to properly join paths when computing libdir/includedir.

[Relevant CMake documentation](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#result-variables)
[More info in the Nix bugtracker](https://github.com/NixOS/nixpkgs/issues/144170)

I ran into that issue while [adding OpenJPH to nixpkgs](https://github.com/NixOS/nixpkgs/pull/324125). For now this commit is maintained as a patch in Nix.


aous72:
Hi Adrian,

Thank you for putting this.
I am happy to merge it. 
Do I need to worry about breaking other repos installation scripts? I am not super familiar what the current trend is.

Thank you.

Kind regards,
Aous.

abustany:
As far as I understand, this change should not break anything. For the record, here's the diff between the old and the new pc file when running `cmake -DCMAKE_INSTALL_PREFIX=/usr`:

```diff
--- openjph.pc.old	2024-07-03 09:08:23.779830131 +0200
+++ openjph.pc	2024-07-03 09:08:40.028517484 +0200
@@ -3,7 +3,7 @@
 Version: 0.14.0
 Requires: 
 prefix=/usr
-includedir=${prefix}/include
-libdir=${prefix}/lib64
+includedir=/usr/include
+libdir=/usr/lib64
 Libs: -L${libdir} -lopenjph
 Cflags: -I${includedir}
```

in this simple case, there's no difference besides the fact that the prefix is now "baked" in includedir/libdir.

The old code breaks however if `CMAKE_INSTALL_INCLUDEDIR` or `CMAKE_INSTALL_INCLUDEDIR` are absolute paths, because `${prefix}` gets prepended unconditionally. The `_FULL` variables, on the other hand, handle those cases correctly.

Nix is probably one of the few cases where those flags are set to absolute paths, Nix uses that to output the files in different prefixes that correspond to different subpackages (one for binaries, one for development headers).

aous72:
Than you Adrien.

aous72:
Hi @kmilos,

Please let me know if Release 0.14.2 works for you.

Also, @abustany, please let me know if Release 0.14.2 works for you.

Kind regards,
Aous.

PR #149, opened 2024-08-02
Merge/aous72 20240802

wayfarer3130:
This is a straight merge from current upstream as of 2024-08-02
This version appears to be much more stable than earlier versions.


aous72:
Hi Bill,

Thank you for this PR.

I am putting my thinking/response on the commit; I am assuming that you intend to use the library for WebAssembly.

Kind regards,
Aous.

src/core/codestream/ojph_codestream_local.cpp
```diff
@@ -74,7 +74,7 @@ namespace ojph {
       cur_comp = 0;
       cur_line = 0;
       cur_tile_row = 0;
-      resilient = false;
```
aous72, line 77: My thinking is that an incomplete file is a catastrophic error.
So the default behavior should be to terminate with an error rather than terminate gracefully.
If the end user accepts this incompleteness, they can use the enable_resilience() function call to enable resilience.

src/core/codestream/ojph_codestream_local.cpp
```diff
@@ -1046,7 +1046,8 @@ namespace ojph {
         int marker_idx = find_marker(infile, next_markers, 2);
         if (marker_idx == -1)
         {
-          OJPH_INFO(0x00030067, "File terminated early");
```
aous72, line 1049: I will get back to you on this one -- I need to check the code.
If you do not want to get any messages, which might be a good idea for WebAssembly, you can use 
`set_message_level(OJPH_MSG_LEVEL::NO_MSG)`

subprojects/js/build.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
 mkdir -p build
-#(cd build && emcmake cmake -DCMAKE_BUILD_TYPE=Debug ..)
-(cd build && emcmake cmake ..)
```
aous72, line 4: This makes the default build is debug.
Is that the intention? can you elaborate?

src/core/codestream/ojph_params.cpp
```diff
@@ -600,45 +600,45 @@ namespace ojph {
...
       if (file->read(&Lsiz, 2) != 2)
-        OJPH_ERROR(0x00050041, "error reading SIZ marker");
+        OJPH_ERROR(0x00050041, "error reading SIZ marker, truncated file");
       Lsiz = swap_byte(Lsiz);
       int num_comps = (Lsiz - 38) / 3;
       if (Lsiz != 38 + 3 * num_comps)
-        OJPH_ERROR(0x00050042, "error in SIZ marker length");
+        OJPH_ERROR(0x00050042, "error in SIZ marker length %d with components %d", Lsiz, num_comps);
       if (file->read(&Rsiz, 2) != 2)
-        OJPH_ERROR(0x00050043, "error reading SIZ marker");
```
aous72, line 609: From this position downwards.
This fine granularity might be useful for a developer; it is of little use to a normal user.
The Rsiz, XSIZ, YSIZ, ... etc. are parameters within the SIZ marker segment.  See the following table.

image

src/core/codestream/ojph_codestream_local.cpp
wayfarer3130: Actually, we do partial decodes all the time for being able to perform streaming decode, and that is enabled by default now, so we always want to allow it.

wayfarer3130:
My apologies - I was trying to merge into a local fork with these changes, not into the master branch.

PR #182, opened 2025-02-19
CMake: polish export config

kmilos:
This is just bringing up the current exported CMake config to "textbook" level:

https://cmake.org/cmake/help/latest/guide/importing-exporting/index.html#exporting-targets
https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html

Should produce the following, more conventional artifacts:
```
cmake/openjph/openjph-config.cmake
cmake/openjph/openjph-config-version.cmake
cmake/openjph/openjph-targets.cmake
cmake/openjph/openjph-targets-<build_type>.cmake
```
@palemieux


aous72:
Hi @kmilos,

Thank you for putting this PR.
I had a look and tried to understand, but it is far more sophisticated than my understanding. 

Kind regards,
Aous.

Issue #165, opened 2024-12-30
I need to expose used_bytes of ojph:::mem_outfile.


aous72:
When a user access compressed data in ```ojph:::mem_outfile```, using ```get_data()```, there is no way to know how many bytes are valid.
```tell()``` can be used, and it is most likely to be correct, but, in reality, it tells where the pointer is -- at the end of compressing a file, ```tell()``` will most like points to one byte beyond the end of the file, and therefore the returned value is actually (without modification) the file size.

A new function call can be useful, or get_data() can be modified to return this -- the first option is better.
```
size_t get_used_bytes() 
{ return used_size; }
```
can do the trick.

aous72:
This is currently can be done by seeking to the end of the file seek ```seek(0, SEEK_END)``` and then using tell().

PR #195, opened 2025-08-25
Exposes used_size for mem_outfile

hmaarrfk:
Closes https://github.com/aous72/OpenJPH/issues/165


hmaarrfk:
is this ok as an implementation/

aous72:
Hi Mark,

Thank you for having a look at my code, and producing this useful commit.
My original thinking was to get to the end of the file, and then use ftell, but this commit is simpler.

I will add it soon, as this requires a version bump.

Kind regards,
Aous.

hmaarrfk:
Awesome thanks. Yeah i was using mem_outfile and figured i would have a look at the open issue around it!

PR #197, opened 2025-08-25
Handle platform diff of -fPIC

### LINKED ISSUE #196: -fPIC was passes to Windows/Clang and caused some errors.
I try to build from source code by Windows/Clang, but -fPIC was added to the compile options.
In src/core/CMakeLists.txt:139:
```cmake
if (NOT MSVC)
  target_compile_options(openjph PRIVATE -fPIC)
endif()
```
It seems that cmake will not set MSVC variable for Clang, and Windows/Clang doesn't support -fPIC.
Maybe change to the following code would be better.
```cmake
# if (NOT MSVC)
#   target_compile_options(openjph PRIVATE -fPIC)
# endif()
set_target_properties(openjph PROPERTIES POSITION_INDEPENDENT_CODE ON)
```

HeaoYe:
Fix #196 

Use cmake higher-level abstractions to handle platform differences of -fPIC
```cmake
# if (NOT MSVC)
#   target_compile_options(openjph PRIVATE -fPIC)
# endif()
set_target_properties(openjph PROPERTIES POSITION_INDEPENDENT_CODE ON)
```

Thanks for considering this contribution!I'm looking forward to your feedback and am open to making any necessary changes.

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -135,9 +135,7 @@ if (BUILD_SHARED_LIBS AND WIN32)
 endif()
 
 ## include library version/name
-if (NOT MSVC)
-  target_compile_options(openjph PRIVATE -fPIC)
-endif()
+set_target_properties(openjph PROPERTIES POSITION_INDEPENDENT_CODE ON)
 target_compile_definitions(openjph PUBLIC _FILE_OFFSET_BITS=64)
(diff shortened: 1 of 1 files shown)


aous72:
Hi HeaoYe,

Thank you for submitting this PR.
Your proposed change is the correct way to have PIC code; I did not know about it.

I will include this PR in the next release.

Kind regards,
Aous.

Issue #198, opened 2025-08-28
aligned memory to 4096


hmaarrfk:
I've found that in many cases aligning memory to 4096 is really beneficial especially when writing large (>1MB) files to disk.

Is this something you would consider?

The implementation might look something like:

<details>

```patch
diff --git a/src/core/others/ojph_file.cpp b/src/core/others/ojph_file.cpp
index d4a4756..359e088 100644
--- a/src/core/others/ojph_file.cpp
+++ b/src/core/others/ojph_file.cpp
@@ -43,6 +43,8 @@
 #include <cassert>
 #include <cstddef>
 
+#include <malloc.h>
+
 #include "ojph_file.h"
 #include "ojph_message.h"
 
@@ -110,7 +112,11 @@ namespace ojph {
   mem_outfile::~mem_outfile()
   {
     if (buf)
+#ifdef OJPH_OS_WINDOWS
+      _aligned_free(buf);
+#else
       free(buf);
+#endif
     is_open = clear_mem = false;
     buf_size = used_size = 0;
     buf = cur_ptr = NULL;
@@ -198,15 +204,26 @@ namespace ojph {
   /** */
   void mem_outfile::expand_storage(size_t needed_size, bool clear_all)
   {
+    auto new_buf = this->buf;
     if (needed_size > buf_size)
     {
       needed_size += (needed_size + 1) >> 1; // x1.5
       si64 used_size = tell(); // current used size
 
-      if (this->buf)
-        this->buf = (ui8*)realloc(this->buf, needed_size);
-      else
-        this->buf = (ui8*)malloc(needed_size);
+#ifdef OJPH_OS_WINDOWS
+      new_buf = (ui8*)_aligned_alloc(4096, needed_size);
+#else
+      new_buf = (ui8*)aligned_alloc(4096, needed_size);
+#endif
+      if (this->buf != new_buf && !clear_all){
+        memcpy(new_buf, this->buf, used_size);
+#ifdef OJPH_OS_WINDOWS
+        _aligned_free(this->buf);
+#else
+        free(this->buf);
+#endif
+      }
+      this->buf = new_buf;
 
       if (clear_mem && !clear_all) // will be cleared later
         memset(this->buf + buf_size, 0, needed_size - this->buf_size);

```
</details>

I'm about to test building on all platforms, but I was wondering if that was something you might want to upstream.

Let me know.

aous72:
Hi Mark,

Thank you for the suggestion.

[...]

PR #200, opened 2025-08-28
Use an aligned memory allocator for the memfile

hmaarrfk:
Tested in https://github.com/conda-forge/openjph-feedstock/pull/11

Closes https://github.com/aous72/OpenJPH/issues/198

### DIFF
--- a/src/core/common/ojph_mem.h
+++ b/src/core/common/ojph_mem.h
@@ -49,23 +49,23 @@
 namespace ojph {
 
   ////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


src/core/others/ojph_file.cpp
```diff
@@ -198,15 +204,26 @@ namespace ojph {
   /** */
   void mem_outfile::expand_storage(size_t needed_size, bool clear_all)
   {
+    auto new_buf = this->buf;
```
hmaarrfk, line 207:
```suggestion
    u8 * new_buf = nullptr;
```

Issue #204, opened 2025-09-15
OpenJPH does not support MSVC aarch64


ggarra13:
Microsoft (R) C/C++ Optimizing Compiler Version 19.44.35213 for ARM64
Copyright (C) Microsoft Corporation.  All rights reserved.

..\OpenEXR-build\_deps\openjph-src\src\core\common\ojph_arch.h(198): fatal error C1189: #error:  Error unsupport MSVC version

ggarra13:
It is not an issue with cmake as it is a clean build.  I am building against v0.22, I believe.

aous72:
Sorry for my incorrect assumption,, Gonzalo, It is indeed an issue with the code.

I did not consider Windows on ARM, and I have no access to such a machine, but I need to fix this issue ASAP.

I am currently on the move on a personal holiday. I will be back next Tuesday.
I will try to test the code with Github Windows on Arm runners, but if that fails II may need your help -- perhaps, I can send you bug fixes, and you would test them for me. It would be great if you have the time.

Kind regards,
Aous.

aous72:
Hi Gonzalo,

Thank you for providing this issue. I hope is solved with the latest update. Let me know if there are any issues.

Feel free to close this issue.

Kind regards,
Aous.

PR #205, opened 2025-09-17
Fixes compilation for Windows on ARM

aous72:
A couple if fixes that makes the code compile and work on Windows on ARM -- I am assuming it is 64bit ARM.


aous72:
Forgot to say this also fixes issue #204

PR #213, opened 2025-09-24
Support testing for both Cygwin and MSYS

kmilos:
[MSYS implies and derives from Cygwin in CMake](https://gitlab.kitware.com/cmake/cmake/-/blob/60cdc353c102804c697f5ee718fe6cb3991ef14f/Modules/Platform/MSYS.cmake)

One could also consider adding [Cygwin to CI](https://github.com/cygwin/cygwin-install-action).


aous72:
Thank you Milos.

PR #214, opened 2025-09-26
'aligned_alloc' requires C++17

noclowns:
Fixing the issue that was causing build errors, at least on macos 10.15.7 
The use of 'aligned_alloc' API requires C++17, by standard.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,9 +118,9 @@ endif()
 message(STATUS "Building ${CMAKE_BUILD_TYPE}")
 
 ## C++ version and flags
-# C++14 is needed for gtest, otherwise, C++11 is sufficient for the library
+## C++14 is needed for gtest, the library requires C++17
 if (NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD 17)
 endif()
 message(STATUS "C++ Standard is set to ${CMAKE_CXX_STANDARD}")
 


aous72:
Hi Slava,

Thank you for putting this, and taking the time to create a PR.

The function `aligned_alloc` is available in C++11; see
https://en.cppreference.com/w/c/memory/aligned_alloc
All my runners in Github Actions run without issues.
I also happen to have a MacBook Air 2012, running 10.15.7, and the code compiles with no issues using clang 12.0 that comes with Xcode.

I suspect, and this is a speculation, that you have `using namespace std` in your code somewhere global, and if I am not wrong, `std::aligned_malloc` is only defined in C++17.

Please let me know if this solves your issue.

Kind regards,
Aous.

noclowns:
Hello Aous. This issue reproduces by simply `brew install openjph` or building it from source via make. This is significant, because there are 391 downstream homebrew formulas with transitive dependency on openjph. 

Link to the repro build log with the failure:
https://gist.github.com/noclowns/f7c39c6757fc5d3ecb9622136b876dbf

### _Build environment:_

- MacOS Catalina 10.15.7
- `Command Line Tools (macOS Mojave version 10.14) for Xcode 10.0` (default that gets installed by`xcode-select --install`). It's not clear why it says "Mojave version 10.14" for Catalina, but could be a significant detail.
- **No** Xcode installed (likely significant because Xcode possibly installs a newer version of SDK headers)

### _Standards discussion:_

You mention _aligned_alloc_ is available since C++11, but note it actually says **C11** (not C++)
https://en.cppreference.com/w/c/memory/aligned_alloc

The _aligned_alloc_ is also available since **C++17** 
https://en.cppreference.com/w/cpp/memory/c/aligned_alloc.html

So we're dealing with a mix of 2 separate standards, one for C and one for C++.
The underlying issue appears to stem from mixed compilation of C and C++ code within a single .cpp file by a C++ compiler (in this case clang++). We're specifying **-std=c++14** to the compiler (< c++17) making _aligned_alloc_ unavailable via "**C++**" standard pathway. And there's no option to, simultaniously, specify the need C-standard level explicitly to the C++ compiler to guarantee the needed function availability via the "**C**" standard pathway. For that a C not C++ compiler must be used. So what is the "C" support level a C++ compiler uses by default?  

In the C++14 standard (ISO/IEC 14882:2014), clause 18.2 “C standard library” defines how the C++ headers i.e. <cstdlib>, map to the C headers. “The contents of these headers are the same as the Standard C library headers as specified in the ISO C standard (ISO/IEC 9899:19**90**)” That is C90, not the later C11 required to implement _aligned_alloc_. 
Also the standard says "Implementations are allowed to expose additional functions", meaning that vendors may expose _aligned_alloc_, but it's implementation-specific. 

It appears that the newer versions of MacOS SDKs are doing just that by tweaking the conditional compilation to include _aligned_alloc_ in the standard headers, but the old versions do not. 

Lets compare the failing older version of the standard headers against the working newer version. Installed MacOS SDKs are located at /Library/Developer/CommandLineTools/SDKs:

_/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk
vs.
/Library/Developer/CommandLineTools/SDKs/MacOSX11.1.sdk_

```
//failing old version
#if (__DARWIN_C_LEVEL >= __DARWIN_C_FULL) **&&** \ //// <===== Note the '&&'
        ((defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L) || \ //// __STDC_VERSION__ must be set for aligned_alloc to be available
         (defined(__cplusplus) && __cplusplus >= 201703L))
void    *aligned_alloc(size_t __alignment, size_t __size) /* */;
#endif

//working newer version
#if (__DARWIN_C_LEVEL >= __DARWIN_C_FULL) **||** \ //// <===== Note the '||' relaxing the condition
    (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L) || \
    (defined(__cplusplus) && __cplusplus >= 201703L)
void    *aligned_alloc(size_t __alignment, size_t __size) /* */;
#endif
```
The difference is that the newer version of MacOS SDK loosens restrictions by not requiring __STDC_VERSION__ to be set to enable aligned_alloc() availability. __STDC_VERSION__ is typically set by the **C** compiler, and is not set in by the **C++** (not C) compiler like clang++. This explains the build failure on the older SDK, although it's not strictly a bug in SDK since it violates no standard.

I was able to bypass the broken build via in-place homebrew formula edits to add "-DCMAKE_CXX_STANDARD=17" to cmake args, and there's various other fixes possible to get the compilation to succeed. 

### _Possible solutions_

- Do nothing. Assuming the impact is limited to a small subset of users/environments, and ad-hoc local fixes are possible. 
- Auto-detect if _aligned_alloc_ is available via Cmake, and dynamically adjust either by bumping up to c++17 (if available) or by conditionally switching to something like _posix_memalign()_ in code. 
- To ensure portability, separate the offending source code into its own .c file to enable specifying  **-std=c11** to the C compiler, and link it back with the rest of the project (proper fix, possible perf impact?)

This was a bit long, but leaving it as a future reference for those who might run into this issue. PR good to close?

aous72:
Thank you a lot Slava.
You comments are very useful. I think they should be available somewhere else, perhaps, on homebrew.

PR #222, opened 2025-10-10
Detect compiler features to enable/disable SIMD support.

patlefort:
This PR check for compiler flags and disable corresponding SIMD support when the flag is not detected.


fundawang:
I think at least neon and sse(including avx etc.) should be distinguished. Because neon does not exsit in x86 targets, the same for SSEses. There is no need build them under different targets.

aous72:
After thinking about it all this time.
I do not think we need it. It complicates the thought process.

PR #242, opened 2026-01-14
Fix: Rule of 3/5/7 violation leads to possible use after free in inadvertent use of copy constructor /assignment.

geo-ant:
First of all, thank you for this fantastic project. I've been using it at work and I've ran into a problem that took me longer to debug than I'm willing to admit :sweat_smile: 

## Context

The problem manifested because I was returning a `mem_outfile` from a function with a signature like this:

```cpp
ojph::mem_outfile encode_j2c_to_mem(uint8_t * data, uint32_t width, uint32_t height) {
   ojph::mem_outfile mf;
   // ... encode image and
   return mf;
}
```

This function could throw, so I was using it in a narrowly scoped try block like so:

```cpp
ojph mem_outfile mf;

try {
   mf = encode_j2c_to_mem(data, width, height);
} catch (std::exception &e) {
   // error handling
}

// proceed using mf
```

What happens here is the following.

## Root Cause

1. the temporary function return value is constructed
2. the copy-assignment operator of instance `mf` is invoked. The compiler will auto-generate this special function for us, which will perform a shallow copy.
3. The temporary is destructed, which will free the pointed-to buffer. That's the problem: the `mf` instance points to the same memory.
4. Once we try to use `mf` for output, we provoke UB, if we're lucky we segfault.

**Godbolt Link**: Here's a demo of the behavior on [Godbolt](https://godbolt.org/z/Y5K1vPKrf).

### Interlude: When it Works

I admit, that the more common use case would be to have the try block enclose the statement, such that we'd write it like so:

```cpp
ojph mem_outfile mf  = encode_j2c_to_mem(data, width, height);
```
and then go on to use `mf`. This won't trigger the use after free. I think because [NRVO](https://en.cppreference.com/w/cpp/language/copy_elision.html) kicks in and the destructor of the temporary isn't called in this case.

## Fixes

To fix this behavior, I did the following:

1. Delete copy constructor and copy assignment. If this exists, it should perform a deep copy. I don't think this needs to exist here.
2. Create move constructor and move assignment. Since the moved-from values should be left in a valid state, I gave them the only valid state that always exists, which is the same as if they were default constructed. This nulls the pointer to the internal buffer, such that the destructor doesn't free the used memory.

### DIFF
--- a/src/core/others/ojph_file.cpp
+++ b/src/core/others/ojph_file.cpp
@@ -96,23 +96,28 @@ namespace ojph {
     fh = NULL;
   }
 
-  //*************************************************************************/
+  ////////////////////////////////////////////////////////////////////////////
+  //
+  //
   // mem_outfile
-  //*************************************************************************/
+  //
+  //
+  ////////////////////////////////////////////////////////////////////////////
 
-  /**  */
+  ////////////////////////////////////////////////////////////////////////////
   mem_outfile::mem_outfile()
   {
     reset(*this);
   }
 
+  ////////////////////////////////////////////////////////////////////////////
   void mem_outfile::reset(mem_outfile& f) noexcept {
     f.is_open = f.clear_mem = false;
     f.buf_size = f.used_size = 0;
     f.buf = f.cur_ptr = nullptr;
   }
 
-  /**  */
+  ////////////////////////////////////////////////////////////////////////////
   mem_outfile::~mem_outfile()
   {
     if (buf)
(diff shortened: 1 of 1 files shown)


src/core/others/ojph_file.cpp
```diff
@@ -117,6 +118,32 @@ namespace ojph {
     buf = cur_ptr = NULL;
   }
 
+  /**  */
```
aous72: In general, I try to avoid many C++ stuff.  
The std::swap here can be replaced by a simple copy because all the types are simple types.
Who knows what the compiler does with these.

Then `rhs` should be initialized to initial state.  For this initial state, I think we need to define a private function `void init()` that initializes all the variables, and call this private function from the default constructor, the move constructor, and the move = operator.

I will do these once I have the time.  Alternatively, you can do them, and let me know.

geo-ant: That sounds sensible to me. I'll take care of it.

src/core/others/ojph_file.cpp
```diff
@@ -117,6 +118,32 @@ namespace ojph {
...
+    std::swap(this->is_open, rhs.is_open);
+    std::swap(this->clear_mem, rhs.clear_mem);
+    std::swap(this->buf_size, rhs.buf_size);
+    std::swap(this->used_size, rhs.used_size);
+    std::swap(this->buf, rhs.buf);
+    std::swap(this->cur_ptr, rhs.cur_ptr);
+  }
+
+  mem_outfile& mem_outfile::operator=(mem_outfile&& rhs) noexcept {
+    // NOTE(geo-ant) this ensures that rhs is in a default-constructed state
```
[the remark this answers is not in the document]
geo-ant: Good catch!

PR #243, opened 2026-01-27
Fix race condition in table initialization with multiple threads

brechtvl:
This caused intermittent wrong encoding in the OpenEXR integration, which then resulted in "Error decoding a codeblock" while decoding.

Use std::call_once to resolve this.


aous72:
I wanted to make changes, but I am short on time.
I think for saving me time and for fixing these issues.

PR #253, opened 2026-02-22
* Validate dimensions in the SIZ marker segment after reading it from a codestream

palemieux:
* Allow fuzzer target to run without a fuzzer available


fuzzing/fuzz_targets/ojph_expand_fuzz_target.cpp
```diff
@@ -94,3 +95,20 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size)
...
+    long len = ftell(f);
+    rewind(f);
+    std::vector<uint8_t> buf(len);
+    fread(buf.data(), 1, len, f);
+    fclose(f);
+    LLVMFuzzerTestOneInput(buf.data(), buf.size());
+  }
+  return 0;
+}
+#endif
```
aous72: add a new line at the end -- it looks like github complains.

src/core/codestream/ojph_params_local.h
```diff
@@ -217,10 +217,37 @@ namespace ojph {
...
+      }
+
+      void set_cod(const param_cod& cod)
       {
         this->cod = &cod;
+      }
 
+      void check_validity()
+      {
         if (XTsiz == 0 && YTsiz == 0)
```
aous72: My only concern is that I modify XTsiz and YTsiz because as I create the codestream, the end user might not have set them.  For reading, my thinking is that they should not be changed,  What do you think?  Perhaps, check_validity should have `bool for_reading`.  It is true, the this check is skipped.

palemieux: Oh. I didn’t notice check_validity() set values!

Instead of for_reading, I would factor out the XTSiz backfill into a dedicated function that is called when writing.

PR #255, opened 2026-02-24
Add NULL checks at tile::pull()

palemieux:
4536793000378368


aous72:
Thank you for putting this.  I think this is a bug somewhere else in the code. I need to get to the bottom of it.

PR #270, opened 2026-05-09
Fix version

sunpoet:
The version update is missing in the latest version (0.27.1).


aous72:
Thank you Po-Chuan Hsieh.

sunpoet:
Thanks!

PR #271, opened 2026-05-11
Add OSS fuzz corpus for ojph_compress_fuzz_target

palemieux:
See https://github.com/google/oss-fuzz/pull/15490


aous72:
Thank you Pierre.

PR #272, opened 2026-05-12
Fix quantizer initialization when QCCs are created

palemieux:
Fixes two related issues:

* `(1 << bit_depth)` overflows when `bit_depth == 32`
* when creating a QCC from the QCD, the `base_delta` of the latter should be used to initialize that of the former.

Also adds a smoke test for the API.


aous72:
The first two files are no longer relevant.
The other two files were move to PR #313.

PR #276, opened 2026-06-02
AVX2 block coder optimizations (encoder + decoder)

osamu620:
## Overview

Draft PR collecting a series of AVX2 optimizations to the HTJ2K block coder
(encoder and decoder), plus supporting architecture-detection and CI changes.
Opening as a draft for early feedback.

Organized into logical commits — generic/arch, AVX2 decoder, AVX2 encoder,
followed by a second round of decoder work (bitstream de-stuffing) and two
fixes (CI runner image, strict-aliasing) — rebased on the current `master`.

## What's included

**AVX2 block decoder** (`ojph_block_decoder_avx2.cpp`, `ojph_block_common.*`)
- Wide UVLC table for non-initial quad rows, combined with a merged dual
  MagSgn fetch and a widened `frwd_advance`; reduced `memset` of the scratch
  buffer.
- Force-inline `decode_two_quad32_avx2` (MSVC-guarded `__forceinline`).
- Decompose the ~770-line `ojph_decode_codeblock_avx2` into a thin dispatcher
  over four `noinline` helpers (`decode_cb_step1_vlc`, `decode_cb_step2_32bit`,
  `decode_cb_step2_16bit`, `decode_cb_spp_mrp`). This isolates each pass's
  register allocation and cuts spilling in the hot step-1 VLC loop.
- **De-stuffed bitstream readers** (second round): each codeblock's MagSgn,
  SPP, VLC, and MRP segments are un-stuffed once into a flat buffer, from
  which bits are fetched at absolute positions (MagSgn/SPP) or through a
  branchless register-resident window (VLC). Since the per-quad bit counts
  depend only on VLC outputs, the loop-carried dependency in the MagSgn pass
  collapses to a single scalar add, and all refill/unstuffing branches leave
  the hot loops. Full-decode branch misses drop ~50–75% and IPC rises from
  ~3.0 to ~3.3–3.5; the stateful `frwd`/`rev` reader machinery is deleted
  from this file. Also replaces the 14-instruction `_mm256_sllv_epi16`
  emulation with a 6-instruction pshufb power-of-two lookup + 16-bit
  multiply.
- Fix pre-existing strict-aliasing violations in this file (type-punned
  `*(ui32*)` accesses in the SPP/MRP and MEL readers, now `memcpy`): GCC
  15/16 on MinGW started miscompiling the SPP pass via TBAA-driven
  dead-store elimination once surrounding code changed. The same punned
  patterns still exist in the other (non-AVX2) block decoder files — see
  #279.

**AVX2 block encoder** (`ojph_block_encoder_avx2.cpp`, `ojph_block_encoder_avx2_apple.h`)
- Branch-overhead reductions in the MEL, MagSgn, and VLC paths.
- Apple-Clang/Intel fallback in a separate header to work around a codegen
  issue observed on that toolchain.

**Architecture / build** (`ojph_arch.*`, CI workflow)
- 64-bit `count_trailing_zeros`, tightened AVX-512 detection, Release CI build.
- CI: `windows-latest` now resolves to the `windows-2025-vs2026` image
  (no Visual Studio 17 2022); the x64 Windows jobs let CMake auto-detect
  the installed Visual Studio instead of hardcoding the generator.

## Correctness & testing

- Decoder output is **byte-identical** to the pre-change build across 8-bit
  (64×64 and 512×8 code-blocks), 16-bit, and lossy inputs, including
  Kakadu-encoded multipass HT codestreams (CUP+SPP+MRP), truncated-stream
  and corrupted-byte cases (identical output or identical error).
- Full `ctest` suite passes (82/82) on Linux/GCC 14 and Windows-MSYS2/GCC 16;
  ASAN and valgrind clean.
- Cross-platform CI is green on the fork: Windows/MSVC, Windows-MSYS2/MinGW,
  macOS Intel + ARM (Apple Clang), Windows-on-ARM, Ubuntu, and EMCC.

## Performance (AMD Zen 5 / GCC, Release, AVX2 tier)

Measured with `-DOJPH_DISABLE_AVX512=ON` so the AVX2 paths are the active
tier. (On AVX-512 hardware the runtime otherwise dispatches to the existing
AVX-512 code in *both* builds, leaving these AVX2 changes dormant; AVX-512
ports of these optimizations are planned as a separate follow-up.) Baseline
is upstream `master`; figures are the median of 11 pinned single-core runs of
the full `ojph_compress` / `ojph_expand`. **Encoded bitstreams and decoded
outputs are byte-identical** between master and this PR — these are pure
performance changes.

| Workload (4K, reversible 5/3 unless noted) | Encode | Decode |
|---|--:|--:|
| 8-bit, 64×64 code-blocks    | **1.91×** | **1.19×** |
| 8-bit, 512×8 code-blocks    | **1.94×** | **1.17×** |
| 8-bit, lossy (qstep 0.005)  | 1.49×     | 1.15×     |
| 16-bit, 64×64 code-blocks   | **1.80×** | **1.29×** |

The encoder gains come from the reduced branch overhead in the MEL / MagSgn /
VLC paths; the decoder gains from the wide UVLC table + merged MagSgn fetch,
the cleanup-pass decomposition, and (second round) the de-stuffed bitstream
readers that remove the serial fetch/advance state and its branches from all
hot loops.

## Status

Draft — opened for early feedback; rebased on the current `master`.

### DIFF
--- a/src/core/coding/ojph_block_common.cpp
+++ b/src/core/coding/ojph_block_common.cpp
@@ -5,7 +5,8 @@
 // Copyright (c) 2022, Aous Naman 
 // Copyright (c) 2022, Kakadu Software Pty Ltd, Australia
 // Copyright (c) 2022, The University of New South Wales, Australia
-// 
+// Copyright (c) 2026, Osamu Watanabe
+//
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions are
 // met:
@@ -32,6 +33,7 @@
 // This file is part of the OpenJPH software implementation.
 // File: ojph_block_common.cpp
 // Author: Aous Naman
+// Author: Osamu Watanabe
 // Date: 13 May 2022
 //***************************************************************************/
 
@@ -102,9 +104,14 @@ namespace ojph {
 
     /// @brief uvlc_tbl0 contains decoding information for initial row of quads
     ui16 uvlc_tbl0[256+64] = { 0 };
-    /// @brief uvlc_tbl1 contains decoding information for non-initial row of 
+    /// @brief uvlc_tbl1 contains decoding information for non-initial row of
     ///        quads
     ui16 uvlc_tbl1[256] = { 0 };
+    /// @brief uvlc_tbl1_wide: wider UVLC table for non-initial rows.
+    ///        Index = mode(2 bits) * 1024 + vlc_data(10 bits) = 12 bits.
+    ///        Entry bits: [4:0]=total_bits, [12:5]=u_q0, [20:13]=u_q1.
+    ///        total_bits == 0x1F means fallback to original decode path.
+    ui32 uvlc_tbl1_wide[4096] = { 0 };
     /// @brief uvlc_bias contains decoding info. for initial row of quads
     ui8 uvlc_bias[256+64] = { 0 };
     /// @}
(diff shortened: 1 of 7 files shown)


src/core/coding/ojph_block_decoder_avx2.cpp
```diff
@@ -784,7 +774,12 @@ namespace ojph {
      *  @param vn       used for handling E values (stores v_n values)
      *  @return __m256i decoded two quads
      */
-    static inline __m256i decode_two_quad32_avx2(__m256i inf_u_q, __m256i U_q, frwd_struct_avx2* magsgn, ui32 p, __m128i& vn) {
+#ifdef OJPH_COMPILER_MSVC
+    static __forceinline
+#else
+    static inline __attribute__((always_inline))
```
palemieux: Does this actually make a noticeable difference?

src/core/coding/ojph_block_common.cpp
palemieux, line 35: Add @osamu620 as author?

src/core/coding/ojph_block_decoder_avx2.cpp
```diff
@@ -898,7 +910,12 @@ namespace ojph {
      *  @return __m128i decoded quad
      */
 
-    static inline __m256i decode_four_quad16(const __m128i inf_u_q, __m128i U_q, frwd_struct_avx2* magsgn, ui32 p, __m128i& vn) {
+#ifdef OJPH_COMPILER_MSVC
```
palemieux: Would it make sense to define a macro for this if-then-else-endif once, since it is used multiple times?

src/core/coding/ojph_block_common.cpp
```diff
@@ -329,6 +334,85 @@ namespace ojph {
...
     }
 
+    //************************************************************************/
+    /** @ingroup uvlc_decoding_tables_grp
+     *  @brief Initializes uvlc_tbl1_wide: wider UVLC table for non-initial
+     *         rows. Index = mode(2b) * 1024 + vlc(10b). Entry packs
+     *         total_bits[4:0], u_q0[12:5], u_q1[20:13].
+     *         total_bits == 0x1F signals fallback to original decode.
+     */
+    static bool uvlc_init_wide_table()
```
palemieux, line 346: @aous72 any reason these tables are generated at runtime instead of being defined statically?

src/core/openjph/ojph_arch.h
```diff
@@ -255,6 +256,35 @@ namespace ojph {
   #endif
   }
 
+  /////////////////////////////////////////////////////////////////////////////
+#ifdef OJPH_COMPILER_MSVC
```
palemieux, line 271: shouldn't this block be moved to the encoder commit since it is used by neither this commit nor the decoder commit.

palemieux (commented):
Nice commit layering!

aous72:
Thank you, @osamu620, for this great contribution. It is like double the encoding performance.
Also, Thanx to @palemieux, for his useful suggestions.

palemieux:
Successfully checked against https://github.com/sandflow/libench for lossless roundtrip.

src/core/coding/ojph_block_decoder_avx2.cpp
osamu620: Yes — it's measurable. Letting `decode_two_quad32_avx2` go out-of-line regressed 16-bit lossless decode by ~8% on Zen 5, since it's called per quad-pair in the hot MagSgn loop (call overhead + lost cross-call register allocation). The 8-bit path uses `decode_four_quad16` the same way. It's now applied via the new `OJPH_FORCE_INLINE` macro.

src/core/coding/ojph_block_common.cpp
osamu620: Done — added `// Author: Osamu Watanabe` to `ojph_block_common.{cpp,h}` (and the copyright headers across the files I touched).

src/core/coding/ojph_block_decoder_avx2.cpp
osamu620: Good idea — added `OJPH_FORCE_INLINE` / `OJPH_NO_INLINE` macros in `ojph_arch.h` and replaced the repeated `#ifdef OJPH_COMPILER_MSVC … #endif` blocks across the encoder and decoder.

src/core/openjph/ojph_arch.h
osamu620: Good catch — moved the 64-bit `count_trailing_zeros` to the encoder commit, since it's only used by the encoder's SWAR MagSgn drain. The arch commit's `ojph_arch.h` change is now just the `OJPH_FORCE_INLINE`/`OJPH_NO_INLINE` macros, which the decoder commit uses.

aous72:
Dear @osamu620,

Whenever you think it is ready to be merge, please let me know.
Thank you for all the effort you put in.

Kind regards,
Aous.

osamu620:
> Dear @osamu620,
> 
> Whenever you think it is ready to be merge, please let me know. Thank you for all the effort you put in.
> 
> Kind regards, Aous.

Dear @aous72

Thank you. Further improvement in decoding speed is being tested in CI/CD.  I will inform you when the PR is ready to be merged.

osamu620:
@aous72 @palemieux 

I feel like that this PR is now ready to be merged. The decoding speed is improved further.

@aous72 

I have found the potential UB in the original code for SigProp decoding. It would be much appreciated if you could check #279

palemieux:
@osamu620 let me know when ready and I will test the PR against the lossless benchmark test set.

osamu620:
> @osamu620 let me know when ready and I will test the PR against the lossless benchmark test set.

@palemieux  I think it's ready.

palemieux (approved):

palemieux:
@osamu620 Passes lossless benchmark round-trip lossless encoding-decoding tests.

PR #282, opened 2026-06-12
Add POWER (VSX) SIMD support for ppc64le (POWER9/POWER10)

runlevel5:
## Summary

This PR adds a native 128-bit VSX SIMD backend for little-endian POWER
(POWER9 / ISA 3.0 and newer), covering the wavelet transform, colour
transform, codestream helper kernels, and the HTJ2K block decoder, with
runtime CPU detection and dispatch. Other PPC targets (POWER8 and
earlier, big-endian) continue to use the generic code paths.

## Design

- **Follows the existing one-file-per-ISA layout**: `ojph_transform_vsx.cpp`,
  `ojph_colour_vsx.cpp`, `ojph_codestream_vsx.cpp`,
  `ojph_block_decoder_vsx.cpp`, with shared 128-bit helpers in
  `ojph_simd_vsx.h`. The kernels are line-for-line ports of the 128-bit
  WASM kernels (the cleanest 128-bit reference), so the two backends are
  easy to diff; there is no source sharing.
- **CPU detection** uses `getauxval(AT_HWCAP/AT_HWCAP2)` on Linux
  (`PPC_FEATURE_HAS_VSX`, `PPC_FEATURE2_ARCH_3_00`,
  `PPC_FEATURE2_ARCH_3_1`), mirroring the existing ARM detection; new
  `PPC_CPU_EXT_LEVEL_*` levels and an `OJPH_ARCH_PPC64LE` define gate
  everything.

Where measurement showed a win, the kernels deviate from a straight port
to use POWER-specific instructions:

- `xvrspi` implements `ojph_round`'s round-to-nearest-ties-away directly,
  replacing a 7-op emulation in the float-to-int conversion kernels
  (emitted via inline asm: GCC's `vec_round` rounds ties to even and
  cannot be used). Verified bit-exact against the scalar reference.
- `vec_sel` replaces 4-op and/andnot/or select idioms in the clipping
  helpers.
- The block decoder destuffs the MagSgn bitstream upfront into a stack
  buffer (same design as the AVX2 decoder), making bit fetches stateless
  and reducing per-quad bit consumption to a GPR add. The stateful
  vector-window approach of the reference decoder defeats store-to-load
  forwarding on POWER and measured ~7% slower than the *scalar* decoder
  on POWER9.

**Block decoder dispatch is content-aware, following the measurements**:
the SIMD decoder is used unconditionally on ISA 3.1 (POWER10), and only
for irreversible tile components on ISA 3.0 (POWER9), where reversible
content (fewer magnitude bits per quad) still slightly favours the
scalar decoder.

## Performance

6144x6144 RGB, single-threaded, GCC 16.1, Fedora 44, `-O3`; baseline is
this repo with `OJPH_DISABLE_SIMD=ON` (note: on ppc64le the baseline ABI
includes VSX, so the baseline already benefits from auto-vectorization).

| Workload   | POWER9 base | POWER9 VSX | POWER10 base | POWER10 VSX |
|------------|-------------|------------|--------------|-------------|
| decode irv | 1.76 s      | **1.39 s (+27%)** | 1.32 s | **0.85 s (+55%)** |
| decode rev | 1.20 s      | 1.20 s     | 0.92 s       | **0.77 s (+19%)** |
| encode     | 2.93/3.38 s | ~+1.5%     | 2.32/2.61 s  | ~+1%        |

Encode gains are small because 82-86% of encode time is in the block
encoder, which this PR does not touch (a candidate for follow-up work).

## Correctness

- Encoded bitstreams (reversible and irreversible) are byte-identical to
  the generic build's on both machines.
- Lossless roundtrip is bit-exact against the source image.
- Decoded outputs are byte-identical to the generic build's in both
  modes (all SIMD paths exercised, including the content-aware decoder
  dispatch on POWER9 and POWER10).
- x86/ARM/macOS/WASM builds are unaffected (all changes are gated on
  `OJPH_ARCH_PPC64LE` / a CMake processor check); macOS arm64 build
  verified clean.

## Limitations / notes for review

- Little-endian only; big-endian PPC falls back to generic code.
- CPU detection is implemented for Linux; other OSes report generic.
- The decoder TU is compiled `-mcpu=power9` and runtime-gated, like the
  AVX512 objects on x86; compiling it a second time at `-mcpu=power10`
  would recover ~2% on POWER10 at the cost of build complexity.
- There is currently no ppc64le CI; a cross-compile + qemu-user job (or
  a Power runner) would be a sensible follow-up.

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -5,11 +5,13 @@ file(GLOB CODESTREAM_SSE2  "codestream/*_sse2.cpp")
 file(GLOB CODESTREAM_AVX   "codestream/*_avx.cpp")
 file(GLOB CODESTREAM_AVX2  "codestream/*_avx2.cpp")
 file(GLOB CODESTREAM_WASM  "codestream/*_wasm.cpp")
+file(GLOB CODESTREAM_VSX   "codestream/*_vsx.cpp")
 file(GLOB CODING           "coding/*.cpp" "coding/*.h")
 file(GLOB CODING_SSSE3     "coding/*_ssse3.cpp")
 file(GLOB CODING_WASM      "coding/*_wasm.cpp")
 file(GLOB CODING_AVX2      "coding/*_avx2.cpp")
 file(GLOB CODING_AVX512    "coding/*_avx512.cpp")
+file(GLOB CODING_VSX       "coding/*_vsx.cpp")
 file(GLOB COMMON           "openjph/*.h")
 file(GLOB OTHERS           "others/*.cpp" "others/*.c")
 file(GLOB TRANSFORM        "transform/*.cpp" "transform/*.h")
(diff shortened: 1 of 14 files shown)


aous72:
Dear [Trung](@runlevel5),

Thank you for this useful contribution.
I tried to add CI to it -- included below -- but I do not have permission to do so.
Perhaps, you can help.

Kind regards,
Aous.



```
name: OpenJPH PowerPC Target Validation

on: [push, pull_request]

jobs:
  test-power-matrix:
    runs-on: ubuntu-24.04-ppc64le  # Leverages native partner-hosted IBM runners
    strategy:
      matrix:
        # Loop through compiler optimization parameters
        target_cpu: [power8, power9, power10] 
    steps:
      - name: Checkout Code
        uses: actions/checkout@v4

      - name: Install Prerequisites
        run: sudo apt-get update && sudo apt-get install -y cmake g++ make

      - name: Configure CMake
        run: cmake -DCMAKE_BUILD_TYPE=Release -DOJPH_ENABLE_TIFF_SUPPORT=OFF -DOJPH_BUILD_TESTS=ON -DCMAKE_CXX_FLAGS="-mcpu=${{ matrix.target_cpu }} -mtune=${{ matrix.target_cpu }}" ..
        working-directory: build

      - name: Compile OpenJPH Source
        run: cmake --build . --config Release
        working-directory: build

      - name: Run Functional Sanity Check
        run: ctest --output-on-failure -C Release
        working-directory: build
```

PR #283, opened 2026-06-12
Fix OJPH_ERROR argument order for NLT/SIZ mismatch message

cary-ilm:
The format string expects SIZ bit depth, signedness, NLT bit depth, signedness, then component index. Passing the component index first caused vfprintf to treat an integer bit depth as a string pointer and crash when reporting the error.


aous72:
Thank you Cary for this PR.
Somehow I did not realize it is a PR; I thought it is just an email. 
I fixed this bug as part of another group of bugs.

PR #285, opened 2026-06-13
AVX2 decoder: N-way codeblock interleave for the step-1 VLC/MEL decode

osamu620:
Follow-up to #276. The first step of decoding a codeblock — recovering the VLC and MEL segments of the HTJ2K cleanup pass (this decoder's "step 1") — is a serial dependency chain and the single hottest decode kernel (~42% of 8-bit AVX2 decode, IPC ~2.8). Codeblocks are independent, so decoding several in lockstep lets the out-of-order engine overlap their chains.

## Two commits

1. **ISA-agnostic plumbing.** That interleaving can't live inside the per-block SIMD decoder, which by definition sees only one codeblock — the code that knows about "a row of codeblocks in a subband" sits one level up, in `subband::pull_line`, so the seam belongs there. This adds an optional `decode_cb32_batch` pointer (next to `decode_cb32`/`decode_cb64`, selected by the same CPU-feature dispatch) and a `codeblock::decode_row` that gathers a row and either calls the batch or falls back to the exact 1-way `decode()` (when no batch is registered, or for 64-bit blocks). **No functional change on its own** — that shared code is touched in just one place, future ISAs plug in with only a kernel + one dispatch line, and current output can't change.

2. **AVX2 kernel.** A templated N-way step-1 kernel (`decode_cb_step1_vlc_nway<N>`) that decodes N same-width codeblocks in lockstep, driven by `ojph_decode_codeblock_avx2_batch`, which groups a subband row's same-width blocks N-at-a-time and falls back to 1-way for edge/odd/invalid blocks. Per-block validation and the second step (MagSgn + optional SgnProp/MagRef) are factored into shared helpers so the 1-way and batched paths stay in sync.

## Performance

Single-thread, `ojph_expand` → tmpfs, `taskset -c0`, best-of-15, vs a 1-way reference build (Zen 5):

| condition | whole-process wall |
|---|---|
| 8-bit lossless | **−9.5%** |
| 16-bit lossless | **−8.6%** |
| 8-bit lossy | **−8.6%** |

Whole-process includes the PPM write, so the decode-kernel gain is larger. `OJPH_DEC_BATCH_N` defaults to 2 — the measured Zen 5 sweet spot (the bulkier AVX2 step-2 between batches makes N=4's larger step-1 working set thrash L1; N=2 also keeps the unrolled per-stream state in registers for 32-bit builds). It's compile-overridable, and it can never regress: anything not batched takes the unchanged 1-way path.

## Correctness

Byte-identical to the 1-way path on 8/16-bit lossless, lossy, the 512×8 streaming shape, and a Kakadu multipass stream (SgnProp/MagRef coverage); 82/82 ctest.

Built clean (0 warnings) locally across GCC on Linux (AVX2-only, AVX-512-enabled, and no-SIMD configs), MSVC VS2022 (x64 + Win32/32-bit), and MinGW UCRT64 GCC 16.

---

@palemieux — would you mind running this against the lossless benchmark round-trip, as you did on #276? Marking draft until that and CI are green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/core/codestream/ojph_codeblock.cpp
+++ b/src/core/codestream/ojph_codeblock.cpp
@@ -223,6 +223,85 @@ namespace ojph {
         zero_block = true;
     }
 
+    //////////////////////////////////////////////////////////////////////////
+    void codeblock::decode_row(codeblock* blocks, ui32 count)
+    {
+      if (count == 0)
+        return;
+
+      // No batched path (the active ISA provides none, or 64-bit codeblocks):
+      // decode each block individually, preserving the exact 1-way behaviour.
+      codeblock_fun& f = blocks[0].codeblock_functions;
+      if (f.decode_cb32_batch == NULL || blocks[0].precision != BUF32)
+      {
+        for (ui32 i = 0; i < count; ++i)
+          blocks[i].decode();
+        return;
+      }
+
+      // Gather the decodable blocks of this row into parallel arrays and hand
+      // them to the batched decoder in windows. All blocks in a subband row
+      // share height (cb_size.h) and stride; only width and the coded data
+      // differ. Zero blocks (no coded passes) are marked directly and left
+      // out of the batch.
+      const ui32 WIN = 64;
+      ui8*  coded[WIN];
+      ui32* dec[WIN];
+      ui32  mmsb[WIN], npass[WIN], len1[WIN], len2[WIN], wid[WIN], orig[WIN];
+      bool  res[WIN];
+
+      const ui32 height = blocks[0].cb_size.h;
+      const ui32 stride = blocks[0].stride;
+      const bool sc     = blocks[0].stripe_causal;
+      ui32 filled = 0;
+
+      for (ui32 i = 0; i <= count; ++i)
+      {
+        if (i < count)
+        {
+          codeblock& cb = blocks[i];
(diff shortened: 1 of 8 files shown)


aous72:
Dear @osamu620,

Thank you for this PR.  
Perhaps, we can talk.

Kind regards,
Aous.

palemieux:
@osamu620  lossless benchmark round-trip passes!

aous72:
Dear @osamu620,

Thank you for all the effort you put in the previous PR.
I know this PR must have taken a lot of effort from you and I am really grateful.
However, this PR put us on a path that we have to reverse in future work; in particular, this PR moves VLC/MEL decoding for two, possibly more, codeblocks to a step by itself, separate or outside from codeblock decoding. This means that we decode VLC/MEL for a second codeblock before decoding MagSgn of the first codeblock; this makes little sense as to why performance has improved.
It is my opinion, with evidence, that it is better to keep all decoding activities for one codeblock within one codeblock decoding unit. I also have evidence that if decoding is optimized properly, we should get a significant speedup of codeblock decoding in the order of 2x to 3x.

I think we should keep this for the time being to see if other parties are interested.

Thank you again.

Kind regards,
Aous

Issue #266, opened 2026-04-11
codestream: I/O operations perform byte swaps independent of machine architecture


SeanMcG:
While attempting to determine test failures with [OpenEXR](http://www.openexr.com/) on a big-endian ppc64 machine (Apple G5), I noticed that it calls in to `openjph` to serialize and deserialize HTJ2K streams but errors out because methods such as `param_siz::write()` performs byte swaps independent of machine architecture, for example from `src/core/codestream/ojph_params.cpp`:

```c++
    bool param_siz::write(outfile_base *file)
    {
      //marker size excluding header
      Lsiz = (ui16)(38 + 3 * Csiz);

      ui8 buf[4];
      bool result = true;

      *(ui16*)buf = JP2K_MARKER::SIZ;
      *(ui16*)buf = swap_byte(*(ui16*)buf);
      result &= file->write(&buf, 2) == 2;
      *(ui16*)buf = swap_byte(Lsiz);
// ...snipped for brevity...
```

SeanMcG:
If I interpret it correctly, The ITU T.814 spec PDF implies the headers are in network byte-order (big-endian) while the stream data itself is not.

michaeldsmith:
yes, the parameter values in marker segments appear in network byte-order (big-endian).  See, e.g., A.1.3 in [JPEG 2000 Part-1 (ITU T.800)](https://www.itu.int/rec/T-REC-T.800-201906-S/en)

<img width="1239" height="1174" alt="Image" src="https://github.com/user-attachments/assets/947462a7-b41b-447a-9c2c-7de486acae6e" />

aous72:
Hi Sean,

Thank you for opening this discussion.
I never payed attention to big endian machines. 
I think there are a lot of issues other than the swap.

Cheers,
Aous.

thesamesam:
>       *(ui16*)buf = swap_byte(*(ui16*)buf);

Please use memcpy for these too, or you'll have more issues like https://github.com/aous72/OpenJPH/issues/186#issuecomment-3801155144.

aous72:
Got it.  Thank you for spotting this again 😃.

michaeldsmith:
@SeanMcG this issue (big-endian machine support) was fixed in commit ecc624e that was merged into master branch and included in the 0.29.0 release

SeanMcG:
Thanks, closing.

PR #287, opened 2026-06-16
Issues/0266 add support for big-endian architecture

michaeldsmith:
* adds support for big-endian architectures 
* fixes issue #266
* add CI that uses big-endian emulation using QEMU
* update Ubuntu Dockerfile
* add Alpine Dockerfile that includes comments in the Dockerfile about how to run QEMU locally using Docker Desktop on Windows to emulate big-endian architecture s390x.


palemieux:
Does the AVX2-accelerated version of `gen_cvrt_32b1c_to_16ub1c_be` at `ojph_img_io_avx2.cpp` need to be modified as well, or does big-endian architecture imply no AVX2 acceleration?

src/apps/others/ojph_img_io.cpp
```diff
@@ -220,15 +220,15 @@ namespace ojph {
       val = *sp0++;
       val = val >= 0 ? val : 0;
       val = val <= max_val ? val : max_val;
-      *p++ = be2le((ui16) val);
+      *p++ = swap_bytes_if_machine_is_little_endian((ui16) val);
```
palemieux:
```suggestion
      *p++ = swap_bytes_if_le((ui16) val);
```

I recommend a shorter function name.

michaeldsmith:
> Does the AVX2-accelerated version of `gen_cvrt_32b1c_to_16ub1c_be` at `ojph_img_io_avx2.cpp` need to be modified as well, or does big-endian architecture imply no AVX2 acceleration?

I think since AVX2 is always on x86 architecture which is little-endian, the machine-dependent switches aren't needed in the AVX2 accelerated version.

src/apps/others/ojph_img_io.cpp
michaeldsmith: I updated the branch to change:
`swap_bytes_if_machine_is_little_endian` to `swap_bytes_if_le`
and
`swap_bytes_if_machine_is_big_endian` to `swap_bytes_if_be`

PR #291, opened 2026-06-17
ppc64le

aous72:
Adds support for powerpc64le.
The original contributor is @runlevel5 in #282.


aous72:
@runlevel5. 
I tested CI on this PR.  
The code does not compile on Ubuntu 22.04 (__builtin_shufflevector was added in gcc 12).
The code compiles and passes the test on Ubuntu 24.04.
The code compiles but fails the test on Ubuntu_latest.
Could you please advise?

Kind regards,
Aous.

aous72:
Now it compiles and works on all of them.

PR #293, opened 2026-06-19
Lock down and clean up the workflows.

dlemstra:
This pull requests locks your workflows down to it's minimal permissions. I also cleaned it a bit up to be more consistent across the workflows. And I add a dependabot.yml to update the actions in your workflows.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    cooldown:
+      default-days: 7
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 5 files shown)


aous72:
Dear Dirk,

Thank you for this PR.
I noticed you have a long string after actions; I suppose it is the commit. For example,
uraimo/run-on-arch-action@f9b26e3a1a408d5fd530d20c17b9f3f4428ff8d9 #3.1.0
Is this to guarantee that no future rogue commit runs on the repository?

Kind regards,
Aous.

dlemstra:
Yeah sorry I should have explained that better in the description. That long string means the versions of the actions are pinned down to a specific commit. And if someone else would take over the repository and change the commit of that version it would not affect you. You can find more information here: https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

PR #297, opened 2026-06-20
Refactor QCC and quantization step size generation

palemieux:


src/core/codestream/ojph_params.cpp
```diff
@@ -1134,149 +1134,85 @@ namespace ojph {
...
+          this->base_delta = 1.0f / (float)(1 << t);
         }
+        this->set_irrev_quant(this->num_decomps);
       }
     }
 
+    //////////////////////////////////////////////////////////////////////////
+    bool param_qcd::is_qcc_needed(ui32 comp_num, const param_cod &cod, const param_siz &siz)
+    {
+      if (! this->is_init)
```
aous72, line 1205: Same as line 1177.
Make the error code 0x00040002.

src/core/codestream/ojph_params.cpp
```diff
@@ -1134,149 +1134,85 @@ namespace ojph {
...
+        qcc->make_quant_steps(c, *coc, siz);
       }
-      else if (other_comps_exist) // Some are captured by QCD
+    }
+
+    //////////////////////////////////////////////////////////////////////////
+    void param_qcd::make_quant_steps(ui32 comp_num, const param_cod &cod, const param_siz &siz)
+    {
+      if (this->is_init)
+        OJPH_ERROR(0x00040001, "Quantization step sizes already initialized.");
```
aous72, line 1177: I know we discuss this before, but I think this should be an assert instead of OJPH_ERROR, as it is internal to the code.

Issue #296, opened 2026-06-20
Failing unit tests when compiling with clang


dlemstra:
While trying to figure out why the build failed (turned out to be gcc version (#295)) I was also testing with clang and saw that a lot of the unit tests failed:

```
59% tests passed, 34 tests failed out of 82

Total Test time (real) =   6.63 sec
```

This can be reproduced with the following `Dockerfile`:

```Dockerfile
FROM ubuntu:24.04

ENV DEBIAN_FRONTEND=noninteractive

RUN apt-get update && apt-get install -y cmake clang git

RUN git clone https://github.com/aous72/OpenJPH

WORKDIR /OpenJPH/build

RUN cmake -DCMAKE_BUILD_TYPE=Release -DOJPH_BUILD_TESTS=ON ..

RUN make
```

aous72:
Hi Dirk,

Thank you for this issue.  I think it should be fixed by the latest commit.
Please reopen it, if the problem was not solved.

Kind regards,
Aous.

dlemstra:
Would you be open to a pull requests so both clang and gcc are tested in the workflow?

aous72:
That would be great. I would be happy to add it.

dlemstra:
I have created this PR: #306

PR #301, opened 2026-06-21
Undefined behaviour fixes in ojph_block_encoder_avx2

aous72:
These bugs were introduced when a new ojph_block_encoder_avx2 was committed.
It was not noticed earlier because it behaved as expected with GCC and MSVC; it did not however with clang.


aous72:
This hopefully addresses #296.

PR #306, opened 2026-06-26
Run the test with both clang and gcc.

dlemstra:


aous72:
That's great.  Thank you Dirk.

PR #308, opened 2026-06-28
Add support for Qfactor

palemieux:
To improve interoperability, the feature uses the visual weights library from @osamu620 at https://github.com/osamu620/OpenHTJ2K/blob/main/source/core/codestream/visual_weighting.hpp, which is copied into OpenJPH.


aous72:
Thank you @palemieux for putting this PR.

src/apps/ojph_compress/ojph_compress.cpp
```diff
@@ -634,6 +641,13 @@ int main(int argc, char * argv[]) {
     return -1;
   }
 
+  if (qfactor != -1 && quantization_step != -1.0f)
+    OJPH_ERROR(0x01000101,
```
aous72: Please change all 0x0100010x to 0x010000Ax.

src/core/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 
-file(GLOB CODESTREAM       "codestream/*.cpp" "codestream/*.h")
+file(GLOB CODESTREAM       "codestream/*.cpp" "codestream/*.h" "codestream/*.hpp")
```
aous72: I would be nicer if we can change the name to .h so it matches the rest of the code; what do you think?

src/core/openjph/ojph_params.h
```diff
@@ -162,6 +162,12 @@ namespace ojph {
   class OJPH_EXPORT param_qcd
   {
   public:
+    enum comp_type {
```
aous72: It would nice to associate a type to the enum, as in
enum comp_type : ui8 {}; or enum comp_type : ui32 {};
If we knew for sure that visual weighting or qfactor will not be assigned to any component other than the first three, we would not need this.

src/core/codestream/ojph_params_local.h
```diff
@@ -711,6 +716,7 @@ namespace ojph {
       bool is_qcc_needed(ui32 comp_num, const param_cod &cod, const param_siz &siz);
       void set_delta(float delta) { base_delta = delta; }
       void set_delta(ui32 comp_idx, float delta);
+      void set_qfactor(ui32 comp_idx, ui8 qfactor, ojph::param_qcd::comp_type ctype);
```
aous72: Is it necessary to pass ojph::param_qcd::comp_type ctype.  We can use the comp_idx for this purpose.  
It unlikely, although possible, that we will use visual weighting for components other than the first 3.
Perhaps, changing the order is useful to 
void set_qfactor(ui32 comp_idx, ojph::param_qcd::comp_type ctype, ui8 qfactor);
What do you think?

src/core/codestream/ojph_params.cpp
```diff
@@ -1289,6 +1314,40 @@ namespace ojph {
       int guard_bits = 1;
       Sqcd = (ui8)((guard_bits<<5)|0x2);//one guard bit, scalar quantization
 
+      // the following are used only when Qfactor is set
+      float qfactor_power = 0;
+      float delta_ref = 0;
+      float G_c = 1;
+      const open_htj2k::visual_weighting_params vp;
+      std::vector<double> W_b;
```
aous72, line 1322: This std::vector objects call new. I would really like to remove them, even if that meant I have to modify Osamu's library.  Ultimately, these are tables with a known number of entries.
The main ojph library have very limited calls to new.
I know Osamu attempts to limit the new calls to one per vector by using reserve.

The horrors of C++.

src/core/codestream/ojph_params_local.h
palemieux: I expect Qfactor may be applied to channels beyond the first three:

* At least in the case of OpenEXR, where images routinely have dozens of channels, with potentially more than 3 containing visual information
* Qfactor can be extended to cover non YCbCr channels by removing/changing the visual weights

Will change the order of the parameters.

src/apps/ojph_compress/ojph_compress.cpp
palemieux: Done

src/core/CMakeLists.txt
palemieux: Done

src/core/openjph/ojph_params.h
palemieux: Done

src/core/codestream/ojph_params_local.h
palemieux: Done

src/core/codestream/ojph_params.cpp
palemieux: Good point.

Questions:
* would you be happy with `reserve()` to avoid multiple calls to new()?
* are you happy with `std::array`?

aous72: I think std::array is a good idea.
For the time being, I will merge the PR, but then I will modify Osamu's code. Do you have any objections?

palemieux: > For the time being, I will merge the PR, but then I will modify Osamu's code. Do you have any objections?

Zero objection assuming that you are ok with submitting a PR against Osamu's library with your proposed improvements.

PR #309, opened 2026-06-30
Fix COC initialization

### LINKED ISSUE #310: Public get_coc() does not initialize the newly created COC
`param_cod::get_coc()` creates a COC if one does not exist using `new param_cod()` through `add_coc_object()`. Unfortunately `new param_cod()` does not initialize the COC values to that of the parent, so that the following ultimately fails:

```
cod.set_reversible(false);

ojph::param_coc coc3 = cod.get_coc(3);
coc3.set_reversible(true);
``` 

P.S.: I would have not expect something called `get_coc()` to modify the state of the underlying object. This is particularly surprising if someone uses when inspecting a decoded codestream. A potential approach: get_coc() would return NULL or error if no COC is available for a particular component; get_effective_coc(comp_idx) would be the effective COC or COD for the component; and get_or_make_coc(comp_idx) would create a COC if one is not available.

palemieux:
Closes #310


palemieux:
Replaced by #312

PR #311, opened 2026-07-01
Improving visual weighting code and implementation.

aous72:
This changes visual weighting in a few ways:
- Replaced std::vector with open_htj2k::visual_weights, which uses std::array<T, 96>; we did this because std::vector needs new/delete, which is unnecessary.
- Replaced double with float because these weights are inaccurate and therefore the precision of double is not needed.
- Historically, when the end user does not specify a quantization step size for irreversible wavelet transform, we set the step size to 2^-bit_depth. In this commit, in lines 1324-1325, we set the quantization step size to 2^-min(16, bit_depth), thus preventing a quantization step size smaller than 2^-bit_depth when the end user does not specify it.

@palemieux Please have a look and let me know if you have any concerns.


src/core/codestream/ojph_visual_weighting.h
```diff
@@ -42,29 +43,64 @@
...
 namespace open_htj2k {
 
+// Maximum number of DWT levels supported.
+constexpr uint8_t max_dwt_levels = 32;
+// Maximum number of visual weight entries (3 subbands per level).
+constexpr size_t max_weight_size = 3 * max_dwt_levels;
+// visual array type
+using visual_weighting_array = std::array<float, max_weight_size>;
+// structure for our visual weights
+class visual_weights {
```
palemieux: FWIW I am not a big fan of this kind of shim that is used only for initialization purposes. I would instead use std::array[] directly.

aous72: Thank you for your comment.
You need to to keep a count of how many entries were used; if not they you rely on the producer and consumer using the same number of entries -- that is the only reason.

palemieux: Ok. Since it is not a generic array, I would

* use more semantically accurate terms, e.g., `count` instead of `size`
* remove `clear()` since it is never used and I cannot imagine a situation where you would want to _c;lear_ visual weights.
* replace `set_weights()` with a constructor
* remove `push_back()`

palemieux: ... actually, why not make the visual weight generation function static members of the `visual_weights` class so that we do not need to expose any accessor methods.

aous72:
@palemieux  I followed you suggestion.  Have a look and let me know if you have any concerns, improvements. Thank you.

PR #312, opened 2026-07-01
Removes direct access to COC segment marker.

aous72:
Direct access to the COC marker segment by the end user of the library proved to be problematic and perhaps confusing.
Here direct access is totally removed.  It is now indirectly accessed through the param_cod object.
This commit borrows code and ideas from PRs #272 and #309.

@palemieux I integrated your  #272 and #309 PRs. I updated the test code.  Have a look and let me know if there are any concerns.  I will merge it tomorrow morning if there are none. Cheers, Aous


tests/test_mixed_coc.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+    codestream.flush();
+    codestream.close();
+  }
+
+  // read back headers and verify per-component settings
+  {
+    ojph::codestream codestream;
+    ojph::j2c_infile j2c_file;
+    j2c_file.open("mixed_rev_irrev_4comp.j2c");
```
palemieux, line 126: This should be created in the build directory, so that it is not checked-in.

palemieux (changes requested):

PR #331, opened 2026-07-28
Cs3d/truncated decode graceful 0.30.1

wayfarer3130:
Fix truncated-codestream exception handling in tile header parsing

## Summary
- Fix exception handling in `tile::parse_tile_header` to also catch `std::exception` and unknown exceptions, not just `const char*`.
- Preserve resilience behavior by logging in resilient mode and only re-raising through `OJPH_ERROR` in non-resilient mode.
- Add a small standalone demo script (`subprojects/js/standalone/truncated_decode_demo.sh`) to reproduce truncated codestream decoding behavior and confirm graceful process termination.

## Problem
When decoding truncated codestreams, OpenJPH may throw `std::runtime_error` via `OJPH_ERROR`.  
`tile::parse_tile_header` currently catches only `const char*`, so `std::runtime_error` bypasses this catch and may terminate the process unexpectedly in downstream WASM/node integrations.

## Root Cause
Mismatch between thrown exception type and caught exception type:
- Throw site: `OJPH_ERROR` -> `std::runtime_error`
- Catch site in `ojph_tile.cpp`: `catch (const char*)`

## Fix Details
In `src/core/codestream/ojph_tile.cpp`, keep the existing handler and add two more:
- `catch (const char* error)` — unchanged; still the path for the `throw "..."`
  string literals in `ojph_bitbuffer_read.h` and `ojph_precinct.cpp`, which are
  the most common truncation errors and carry useful messages.
- `catch (const std::exception& error)` — new; covers `std::runtime_error` from
  `OJPH_ERROR` and `std::bad_alloc` from the allocators.
- `catch (...)` — new; backstop so nothing escapes `parse_tile_header`.

## Why This Is Safe
- No behavior change on successful decodes.
- The `const char*` handler is untouched, so existing truncation diagnostics keep
  their original messages.
- In resilient mode, errors continue to be reported via `OJPH_INFO`.
- In non-resilient mode, failures still propagate as errors through `OJPH_ERROR`.

## Repro / Validation
1. Start from a valid `.j2c`.
2. Truncate to first 10 KiB.
3. Decode with `ojph_expand`.
4. Verify process exits normally (possibly non-zero), instead of aborting via uncaught exception.

Standalone script:
- `subprojects/js/standalone/truncated_decode_demo.sh`

## Notes
Downstream WASM wrappers may also require exception-catching support at link time to avoid runtime aborts when exceptions are thrown across wrapper boundaries.


aous72:
Hi Bill, Alireza,

Thank you for the wonderful cornerstone tools; I know you were early adopters of OpenJPH tool.
Thank you for this PR.

Sorry, I should have written to you earlier.
These are my thoughts:
* I do not think making resilient the default mode is a good idea, as most user need to know if the codestream is broken; this is however not your usual use case.
* If info is annoying you by polluting the webbrowser log, you can always remove the info  by calling `set_message_level` with `OJPH_MSG_WARN` to remove info messages.
* As for the exception throwing, my original design was to throw only `char *` from the decode path (I know char * is a bad idea). However, this might no longer be the case; I need to check that.  
  * Also, my executable, `ojph_expand`, runs the library code inside try-catch block to catch any throws; is this broken now? I need to check. Perhaps, your code can also catch any throws from my library code.
* It is good idea to have this test code added to the repository. However, the code is a bash shell which does not work in Windows.  I am happy to write a corresponding C++ code for the test, and added to the test build.

Kind regards,
Aous

wayfarer3130:
Hi Aous,

I'd still like to see the truncated code stream work.  What about having this throw by default, but using the flag to allow proceeding on warnings to enable truncated decode?  The message can be managed by setting to WARN as you suggest.

Thank you,
Bill

aous72:
I would also like to see the truncated code stream work.
Is the problem addressed by this PR new, or is it an old problem? I know you have been using this library for some time -- no need for testing, just tell me what you know. I am just trying to better understand the problem.

Sorry, I will be going to bed soon.

wayfarer3130:
The problem addressed is quite old - you CAN get truncated code streams to work if you know/provide a full buffer for the full original size, but if you don't know the original size then it will fail.  It also occurs on quite small lossy encoded images even when the entire buffer is available.  The original tests for decoding sub-parts used known size buffers with partial data.

wayfarer3130:
@aous72 - I've addressed your comments.  Thank you for taking time to look at this.  I think the current version is closer to keeping the original design, and it works quite well for Cornerstone usage since the resilient flag can simply be set on call when decoding a partial stream.
I've added a test so that you can run it and see the issue - the test will fail against the previous code version - it is only error handling correctness, but it has been there a while.

aous72:
Thanks a lot for this update.

PR #336, opened 2026-07-29
Fix: raw_in reads every .raw file as signed, and zero-extends signed 16-bit samples

### LINKED ISSUE #335: ojph_compress raw input: signed 2-byte samples are zero-extended (negatives corrupt on encode)
## Summary

`ojph_compress` fed a raw/yuv 16-bit **signed** input file (`-signed true
-bit_depth 16`) silently corrupts every sample with a negative value on
encode — the resulting codestream is wrong, not just misread. Every
negative `int16` value comes back as `INT16_MIN` (`0x8000`) regardless of
its original magnitude. Non-negative samples are unaffected. Reproduces
identically via `ojph_expand` (this project's own decoder) and via
`imagecodecs` (a separate, OpenJPEG-derived decoder) — so this is not a
decoder-side artifact, and not specific to any one downstream consumer.

Regression window: **`0.28.1` (correct) → `0.29.0` (broken)**, and every
release since (`0.30.0`, `0.30.1`, `0.31.0`, the latest at time of filing)
reproduces the same signature.

## Minimal repro

```sh
python3 - <<'PY'
import numpy as np
np.array([[100, -1], [200, -32768]], dtype="<i2").tofile("f.raw")
PY

ojph_compress -i f.raw -o f.j2c \
  -dims '{2,2}' -num_comps 1 -signed true -bit_depth 16 -downsamp '{1,1}' \
  -prog_order RPCL -num_decomps 5 -reversible true

ojph_expand -i f.j2c -o f_out.raw

python3 - <<'PY'
import numpy as np
print("in :", np.fromfile("f.raw", dtype="<i2"))
print("out:", np.fromfile("f_out.raw", dtype="<i2"))
PY
```

**Expected:** `out == in == [100, -1, 200, -32768]`

**Actual:** `out == [100, -32768, 200, -32768]`.

An exhaustive check — a 256×256 frame containing all 65536 possible
`int16` values exactly once — shows exactly the negative half is
corrupted: **32768/65536 mismatches**, every one of them a negative input
value, every one of them decoded back as `-32768`.

## Version matrix (bisected: `git checkout <tag>`, `cmake
-DBUILD_SHARED_LIBS=OFF -DOJPH_ENABLE_TIFF_SUPPORT=OFF`, macOS arm64,
AppleClang, all tags built from a clean checkout)

| Tag | Result |
|---|---|
| 0.26.3 | correct (full int16 range tested) |
| 0.27.0 | correct |
| 0.28.1 | correct (also re-verified exhaustively — see below) |
| **0.29.0** | **broken** |
| 0.30.0 | broken |
| 0.30.1 | broken |
| 0.31.0 | broken |

No tags exist between `0.28.1` and `0.29.0`, so the regression window is
exactly that release.

## Root cause

`git diff 0.28.1 0.29.0 -- src/apps/others/ojph_img_io.cpp`, function
`raw_in::read()`, the `bytes_per_sample > 1` (9–16 bit) `is_signed`
branch:

```diff
       if (is_signed) {
         const si16* sp = (si16*)buffer;
         for (ui32 i = width; i > 0; --i, ++sp)
-          *dp++ = *sp;
+          *dp++ = swap_bytes_if_be((ui16)*sp);
```

(`dp` is `si32*`, the internal line buffer.) `swap_bytes_if_be(ui16)`
(`src/core/openjph/ojph_arch.h:388`) is a no-op on little-endian machines
and returns `ui16`. Casting `*sp` (signed) to `ui16` first discards its
sign; the subsequent implicit `ui16 -> si32` conversion on assignment to
`*dp` **zero-extends** rather than **sign-extends**: `*sp == -1` (bit
pattern `0xFFFF`) becomes `(ui16)0xFFFF == 65535`, then `*dp = 65535`
instead of `-1`. This landed alongside the same release's big-endian/S390x
byte-swap support ("S390x Support (#290)").

Two adjacent branches in the same function were not affected, which is a
useful comparison:
- `bytes_per_sample > 3` (32-bit) `is_signed` keeps an explicit outer
  cast: `*dp++ = (si32)swap_bytes_if_be((ui32)*sp);` — correctly
  re-interprets the swapped bits as signed at their own width before
  assignment (no widening happens, so no extension is needed).
- `bytes_per_sample > 2` (24-bit) `is_signed` manually sign-extends:
  `val |= (val & 0x800000) ? (si32)0xFF000000 : 0;`.

Only the 16-bit `is_signed` branch lost sign-extension when its byte-swap
wrapper was introduced — it widens `ui16 -> si32` without first
recovering the sign at the 16-bit width.

## Suggested fix (we vendor this locally, see below)

```diff
       if (is_signed) {
         const si16* sp = (si16*)buffer;
         for (ui32 i = width; i > 0; --i, ++sp)
-          *dp++ = swap_bytes_if_be((ui16)*sp);
+          *dp++ = (si32)(si16)swap_bytes_if_be((ui16)*sp);
```

Reinterpret the (possibly byte-swapped) bits as `si16` — recovering the
sign at the 16-bit width, mirroring what the 32-bit branch already does —
before widening to `si32`.

**Verified**: a build with this one-line change round-trips ALL 65536
possible `int16` values bit-exact (natural order and shuffled order, via
both the patched `ojph_expand` and `imagecodecs`), where the unpatched
build fails exactly the negative half. The unsigned path (re-verified with
the patched binary) is unaffected. See the companion PR draft
(`upstream-pr-draft.md`, same directory) for the full patch and commands.

## Environment

- macOS arm64 (Apple M-series), AppleClang 21.0.0
- Built via `cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF
  -DOJPH_ENABLE_TIFF_SUPPORT=OFF`
- Confirmed independent of decoder: reproduces via OpenJPH's own
  `ojph_expand` (same build) AND via `imagecodecs` 2026.6.26 (bundled
  OpenJPEG-derived JPEG2000 decoder)
- Not yet tested on x86_64/Linux or a big-endian target (ironically the
  target the byte-swap refactor was for — we have not verified whether
  big-endian hosts are affected differently or the same way)

## Impact / why we filed

We hit this pinning `ojph_compress -signed true` for lossless archival
encoding of DICOM medical images with `PixelRepresentation=1` (signed
pixel data, common in MR/PT). Any signed series containing a genuinely
negative stored pixel value is silently archived as corrupt data with
`0.29.0` onward — no error, no warning, wrong pixels, discoverable only by
diffing against the source. We are vendoring the one-line patch above
onto our pinned `0.31.0` build until this ships upstream, and will retest
against any official fix.

curtisc:
Fixes #335 (signed 2-byte raw samples zero-extended on encode — every negative 16-bit signed raw input sample corrupts, regression window 0.28.1 → 0.29.0).

Two defects in the `.raw` input path. They are entangled: fixing either one on its own leaves the other kind of `.raw` input corrupt, which is why this PR grew past the one-liner it started as. Tests covering all eight sample width / signedness combinations are included, and the whole thing is verified on a big-endian host.

### 1. `raw_in::read()` zero-extends signed 16-bit samples

The `bytes_per_sample > 1` (9–16 bit) `is_signed` branch changed in 0.29.0, alongside that release's big-endian/S390x byte-swap support, from:
```cpp
*dp++ = *sp;
```
to:
```cpp
*dp++ = swap_bytes_if_be((ui16)*sp);
```
`swap_bytes_if_be(ui16)` returns `ui16`; the implicit `ui16 → si32` conversion on assignment to `dp` (`si32*`) zero-extends instead of sign-extending. The codestream is therefore encoded wrong (not misread later) — it reproduces identically via `ojph_expand` and third-party decoders (imagecodecs/OpenJPEG).

Fixed by reinterpreting the (possibly byte-swapped) bits as `si16` — recovering the sign at the 16-bit width — before widening. This mirrors the `bytes_per_sample > 3` (32-bit) `is_signed` branch's explicit outer cast (`(si32)swap_bytes_if_be((ui32)*sp)`), which was not affected, and it mirrors `raw_out::write()`'s 16-bit signed branch, which already writes `(si16)swap_bytes_if_be((ui16)(si16)val)`. The reader was the only half of that pair missing the cast. The 24-bit branch sign-extends manually and is likewise unaffected, as are all `is_signed == false` branches, which correctly want zero extension.

### 2. `raw_in` is never told the signedness that `-signed` requested

`ojph_compress`'s `is_signed` is a `si32*` holding one entry per component, and it was passed straight to:
```cpp
void raw_in::set_img_props(const size& s, ui32 bit_depth, bool is_signed);
```
The pointer converts to `bool`, which is always `true`, so **every** `.raw` file is read through the reader's *signed* branches whatever `-signed` said. Fixed by passing `is_signed[0] == 1`, the way the `.yuv` branch already reads that same array.

### Why both are needed

With the reader locked in its signed branch by (2), the zero extension of (1) was also exactly what kept unsigned 16-bit input intact. Fixing (1) alone moves the corruption from signed input to unsigned input; fixing (2) alone leaves #335 unfixed. Fixing (2) additionally repairs unsigned 8-bit and unsigned 24-bit `.raw` input, which are broken on master today for the same reason: the wrongly chosen signed branch sign-extends any sample whose top bit is set, and `raw_out::write()` then clamps the resulting negative value to 0.

`-reversible true`, one 256x256 component, decoded frame compared to the input sample by sample. Identical results on little- and big-endian:

| `-bit_depth` / `-signed` | master (`b44e3d8`) | commit 1 only | this PR |
|---|---|---|---|
| 8 signed | pass | pass | pass |
| 8 unsigned | **fail** | **fail** | pass |
| 16 signed | **fail** (#335) | pass | pass |
| 16 unsigned | pass | **fail** | pass |
| 24 signed | pass | pass | pass |
| 24 unsigned | **fail** | **fail** | pass |
| 32 signed | pass | pass | pass |
| 32 unsigned | pass | pass | pass |

32-bit is unaffected in both directions because that branch's explicit `(si32)` outer cast makes the signed and unsigned reads produce identical bit patterns.

### Tests

`tests/test_executables.cpp` gains eight `.raw` round trips, one per sample width and signedness — the eight combinations `raw_in::read()` and `raw_out::write()` each branch on, none of which had coverage before (the existing raw tests all go through `yuv_in` with 8-bit unsigned samples).

Each writes one 256x256 single-component frame, encodes it reversibly, decodes it, and requires the result to be bit exact. The frame ramps from the smallest to the largest sample the component can hold, so at 16 bits the step is exactly 1 and one frame sweeps all 65536 sample values: the negative half of the range, where the reader regressed, is covered exhaustively rather than sampled. Sample bytes are composed and read back with explicit shifts, so the fixture and the comparison are byte-identical on either endianness and neither depends on the code under test.

Against `b44e3d8` these tests fail for 8-bit unsigned, 16-bit signed and 24-bit unsigned samples.

### Verification

- **Little-endian** — macOS arm64, AppleClang: `ctest` 94/94, no new compiler warnings under `-Wall -Wextra -Wconversion`.
- **Big-endian** — s390x, Ubuntu 22.04, g++ 11.4.0, under the same QEMU emulation the `test_s390x` job uses, with `byte order: BIG ENDIAN` asserted in-process at the start of the run: `ctest` 94/94 (103 s).
- The same eight tests against `b44e3d8`'s apps fail for 8-bit unsigned, 16-bit signed and 24-bit unsigned on **both** hosts, with byte-identical failure signatures — same mismatch counts, same first differing sample, same values. That is what a defect in the widening rather than in the swap should look like; the fix composes with `swap_bytes_if_be()` instead of bypassing it.
- #335's minimal repro round trips bit exact: `(100, -1, 200, -32768)`.
- Exhaustive 16-bit signed sweep, all 65536 values, checked through `ojph_expand` and independently through imagecodecs: 0 mismatches.

### DIFF
--- a/src/apps/ojph_compress/ojph_compress.cpp
+++ b/src/apps/ojph_compress/ojph_compress.cpp
@@ -1049,7 +1049,7 @@ int main(int argc, char * argv[]) {
           OJPH_ERROR(0x01000085,
             "-downsamp option is missing and must be provided\n");
 
-        raw.set_img_props(dims, bit_depth[0], is_signed);
+        raw.set_img_props(dims, bit_depth[0], is_signed[0] == 1);
 
         siz.set_num_components(num_components);
         siz.set_component(0, comp_downsampling[0], bit_depth[0], is_signed[0]);
--- a/src/apps/others/ojph_img_io.cpp
+++ b/src/apps/others/ojph_img_io.cpp
@@ -1586,7 +1586,7 @@ namespace ojph {
       if (is_signed) {
         const si16* sp = (si16*)buffer;
         for (ui32 i = width; i > 0; --i, ++sp)
-          *dp++ = swap_bytes_if_be((ui16)*sp);
+          *dp++ = (si32)(si16)swap_bytes_if_be((ui16)*sp);
       }
       else {
         const ui16* sp = (ui16*)buffer;
--- a/src/apps/others/ojph_img_io.cpp
+++ b/src/apps/others/ojph_img_io.cpp
@@ -1591,7 +1591,7 @@ namespace ojph {
       else {
         const ui16* sp = (ui16*)buffer;
         for (ui32 i = width; i > 0; --i, ++sp)
-          *dp++ = swap_bytes_if_be(*sp);
+          *dp++ = (si32)swap_bytes_if_be(*sp);
       }
     }
     else
--- a/src/core/shared/ojph_simd_vsx.h
+++ b/src/core/shared/ojph_simd_vsx.h
@@ -266,8 +266,10 @@ static inline v128_t vsx_i64x2_extend_low_i32x4(v128_t a)
   // return (v128_t)__builtin_convertvector(
   //   __builtin_shufflevector(v, v, 0, 1), vsx_v_i64);
 
-  // Unpacks and sign-extends elements 0 and 1 on Little Endian
-  return (v128_t)vec_unpackl((vsx_v_i32)a);
+  // vec_unpackh's "high" half is the one holding elements 0 and 1; the
+  // compiler keeps that mapping on both endiannesses, so this sign-extends
+  // elements 0 and 1.
+  return (v128_t)vec_unpackh((vsx_v_i32)a);
 }
 static inline v128_t vsx_i64x2_extend_high_i32x4(v128_t a)
 {
@@ -275,8 +277,10 @@ static inline v128_t vsx_i64x2_extend_high_i32x4(v128_t a)
   // return (v128_t)__builtin_convertvector(
   //   __builtin_shufflevector(v, v, 2, 3), vsx_v_i64);
 
-  // Unpacks and sign-extends elements 2 and 3 on Little Endian
-  return (v128_t)vec_unpackh((vsx_v_i32)a);
+  // vec_unpackl's "low" half is the one holding elements 2 and 3; the
+  // compiler keeps that mapping on both endiannesses, so this sign-extends
+  // elements 2 and 3.
+  return (v128_t)vec_unpackl((vsx_v_i32)a);
 }
 
 //---------------------------------------------------------------------------
--- a/tests/test_executables.cpp
+++ b/tests/test_executables.cpp
@@ -257,6 +257,174 @@ void compare_files(const std::string& base_filename,
 }
 
 ////////////////////////////////////////////////////////////////////////////////
+//                          run_ojph_compress_raw
+////////////////////////////////////////////////////////////////////////////////
+// Same as run_ojph_compress, but the input file is one the test generated in
+// OUT_FILE_DIR rather than one of the reference files.
+void run_ojph_compress_raw(const std::string& src_filename,
+  const std::string& base_filename,
+  const std::string& out_ext,
+  const std::string& extra_options)
+{
+  try {
+    std::string result, command;
+    command = std::string(COMPRESS_EXECUTABLE)
+      + " -i " + OUT_FILE_DIR + src_filename
+      + " -o " + OUT_FILE_DIR + base_filename + "." + out_ext
+      + " " + extra_options;
+    EXPECT_EQ(execute(command, result), 0);
+  }
+  catch (const std::runtime_error& error) {
+    FAIL() << error.what();
+  }
+}
+
+////////////////////////////////////////////////////////////////////////////////
+//                             raw_test_sample
+////////////////////////////////////////////////////////////////////////////////
+// The .raw tests below all use one frame shape, RAW_TEST_WIDTH by
+// RAW_TEST_HEIGHT, and fill it with a ramp that starts at the smallest and
+// ends at the largest sample the component can hold.  The two dimensions
+// multiply to 65536, so for a 16 bit component the ramp's step is exactly 1
+// and one frame sweeps every value a 16 bit sample can take, exactly once;
+// for the other widths the ramp covers the whole range in even steps, and the
+// two ends of the range are always included.
+#define RAW_TEST_WIDTH        256
+#define RAW_TEST_HEIGHT       256
+#define RAW_TEST_NUM_SAMPLES  (RAW_TEST_WIDTH * RAW_TEST_HEIGHT)
+
+static
+ojph::si64 raw_test_sample(ojph::ui32 idx, ojph::ui32 bit_depth,
+                           bool is_signed)
+{
+  ojph::si64 lower, upper;
+  if (is_signed) {
+    upper = ((ojph::si64)1 << (bit_depth - 1)) - 1;
+    lower = -((ojph::si64)1 << (bit_depth - 1));
+  }
+  else {
+    upper = ((ojph::si64)1 << bit_depth) - 1;
+    lower = 0;
+  }
+  return lower
+    + (ojph::si64)idx * (upper - lower) / (RAW_TEST_NUM_SAMPLES - 1);
+}
+
+////////////////////////////////////////////////////////////////////////////////
+//                           write_raw_test_file
+////////////////////////////////////////////////////////////////////////////////
+// Writes the frame described above to OUT_FILE_DIR as a one component .raw
+// file.  .raw files hold samples least significant byte first; the bytes are
+// composed with shifts rather than by copying a wider integer, so that the
+// file this writes is identical on little- and big-endian machines.
+void write_raw_test_file(const std::string& filename, ojph::ui32 bit_depth,
+  bool is_signed)
+{
+  std::string name = std::string(OUT_FILE_DIR) + filename;
+  FILE* f = fopen(name.c_str(), "wb");
+  if (f == NULL) {
+    FAIL() << "Unable to open file " << name << " for writing.";
+    return;
+  }
+
+  ojph::ui32 bytes_per_sample = (bit_depth + 7) >> 3;
+  ojph::ui8 line[RAW_TEST_WIDTH * 4];
+  size_t line_size = (size_t)RAW_TEST_WIDTH * bytes_per_sample;
+  for (ojph::ui32 y = 0; y < RAW_TEST_HEIGHT; ++y) {
+    ojph::ui8* dp = line;
+    for (ojph::ui32 x = 0; x < RAW_TEST_WIDTH; ++x) {
+      ojph::ui32 idx = y * RAW_TEST_WIDTH + x;
+      ojph::ui64 val = (ojph::ui64)raw_test_sample(idx, bit_depth, is_signed);
+      for (ojph::ui32 b = 0; b < bytes_per_sample; ++b)
+        *dp++ = (ojph::ui8)((val >> (8 * b)) & 0xFFu);
+    }
+    EXPECT_EQ(fwrite(line, 1, line_size, f), line_size);
+  }
+  fclose(f);
+}
+
+////////////////////////////////////////////////////////////////////////////////
+//                          compare_raw_test_file
+////////////////////////////////////////////////////////////////////////////////
+// Checks a decoded one component .raw file against the frame that
+// write_raw_test_file() wrote, sample by sample.  The sign of a negative
+// sample is recovered with an explicit subtraction, so that this reads the
+// file the same way on little- and big-endian machines.
+void compare_raw_test_file(const std::string& filename, ojph::ui32 bit_depth,
+  bool is_signed)
+{
+  std::string name = std::string(OUT_FILE_DIR) + filename;
+  FILE* f = fopen(name.c_str(), "rb");
+  if (f == NULL) {
+    FAIL() << "Unable to open file " << name << " for reading.";
+    return;
+  }
+
+  ojph::ui32 bytes_per_sample = (bit_depth + 7) >> 3;
+  ojph::ui64 sign_bit = (ojph::ui64)1 << (8 * bytes_per_sample - 1);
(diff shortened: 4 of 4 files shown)


curtisc:
## Big-endian results, tests — and the one-line fix on its own is not safe to merge

Two things to report: the big-endian verification that was missing, and a problem the new tests found in my original patch.

### The original one-line fix regresses unsigned 16-bit `.raw` input

`raw_in` is never told what `-signed` said. `ojph_compress`'s `is_signed` is a `si32*` holding one entry per component, and it was handed straight to

```cpp
void raw_in::set_img_props(const size& s, ui32 bit_depth, bool is_signed);
```

The pointer converts to `bool`, which is always `true`, so **every** `.raw` file is read through the reader's *signed* branches regardless of `-signed`.

That is what made the one-line patch look complete: with the reader stuck in the signed 16-bit branch, the zero extension that corrupted signed samples was also exactly what kept unsigned samples intact. Correcting the extension without correcting the signedness moves the corruption from signed input to unsigned input — it does not remove it.

The second commit passes `is_signed[0] == 1`, the way the `.yuv` branch already reads that same array (`ojph_compress.cpp:989`). That also repairs unsigned 8-bit and unsigned 24-bit `.raw` input, which are broken on master today for the same reason: the wrongly chosen signed branch sign-extends every sample whose top bit is set, and `raw_out::write()` then clamps the resulting negative value to 0.

If you would rather have that as its own PR against an issue of its own, say so and I will split it — but the two cannot land separately without a window in which one kind of `.raw` input is corrupt.

### Round-trip status

`-reversible true`, one 256x256 component, decoded frame compared to the input sample by sample. Identical results on little- and big-endian.

| `-bit_depth` / `-signed` | master (`b44e3d8`) | first commit only | this PR |
|---|---|---|---|
| 8 signed | pass | pass | pass |
| 8 unsigned | **fail** | **fail** | pass |
| 16 signed | **fail** (#335) | pass | pass |
| 16 unsigned | pass | **fail** | pass |
| 24 signed | pass | pass | pass |
| 24 unsigned | **fail** | **fail** | pass |
| 32 signed | pass | pass | pass |
| 32 unsigned | pass | pass | pass |

32-bit is unaffected in both directions because that branch's explicit `(si32)` outer cast makes the signed and unsigned reads produce identical bit patterns.

### Big-endian verification

Reproduced your `test_s390x` job locally — s390x, Ubuntu 22.04, g++ 11.4.0, under the same QEMU emulation `uraimo/run-on-arch-action` uses (no native s390x hardware here). The endianness is asserted in-process at the start of the run rather than assumed:

```
$ uname -m
s390x
byte order: BIG ENDIAN
g++ (Ubuntu 11.4.0-1ubuntu1~22.04.3) 11.4.0

$ cmake -DCMAKE_BUILD_TYPE=Release -DOJPH_BUILD_STREAM_EXPAND=ON \
        -DOJPH_ENABLE_TIFF_SUPPORT=ON -DOJPH_BUILD_TESTS=ON .. && make && ctest
100% tests passed, 0 tests failed out of 94
Total Test time (real) = 103.15 sec
```

Then the same test binary against apps built from `b44e3d8` on that same big-endian host:

```
[  FAILED  ] TestExecutables.SimpleEncRev53Raw8Unsigned
  32640 of 65536 samples do not survive the round trip;
  the first is sample 32896, which should be 128 but is 0.
[  FAILED  ] TestExecutables.SimpleEncRev53Raw16Signed
  32767 of 65536 samples do not survive the round trip;
  the first is sample 1, which should be -32767 but is -32768.
[  FAILED  ] TestExecutables.SimpleEncRev53Raw24Unsigned
  32768 of 65536 samples do not survive the round trip;
  the first is sample 32768, which should be 8388735 but is 0.
```

Those failures are byte-identical to the little-endian ones — same mismatch counts, same first differing sample, same values. That is what you would expect of a defect in the widening rather than in the swap: the fix composes with `swap_bytes_if_be()` (swap at 16 bits, reinterpret as `si16`, then widen) instead of bypassing it, so it behaves the same in both byte orders.

### Tests

`tests/test_executables.cpp` gains eight `.raw` round trips, one per sample width and signedness — the eight combinations `raw_in::read()` and `raw_out::write()` each branch on, none of which had coverage (the existing raw tests all go through `yuv_in` with 8-bit unsigned samples).

Each test writes one 256x256 single-component frame, encodes it reversibly, decodes it, and requires the result to be bit exact. The frame ramps from the smallest to the largest sample the component can hold, so at 16 bits the step is exactly 1 and **one frame sweeps all 65536 sample values** — the negative half of the range, where the reader regressed, is covered exhaustively rather than sampled.

Two notes on how they are written, both easy to change if you would rather they were shaped differently:

- **The input files are generated by the test rather than fetched.** Adding them to `jp2k_test_codestreams` would have been more consistent with the rest of the file, but that is your repository; generating them also keeps an exhaustive 16-bit sweep out of the fixture download. The bytes are composed with explicit shifts, so the generated file and the comparison are byte-identical on either endianness and neither depends on the code under test.
- **The comparison is in-process rather than via `mse_pae` or `compare_files`.** `mse_pae` understands signed 16-bit `.rawl` input, but only under a filename ending in `.rawl`, which `ojph_expand` will not write; `compare_files` skips JPEG 2000 comment segments, which is not meaningful over pixel data. The in-process check reports the first differing sample, which is what made the regression above obvious.

While iterating I hit one thing worth mentioning: `tests/ojph_compress` and `tests/ojph_expand` are refreshed by a `POST_BUILD` command on `test_executables`, so after changing only `src/apps/**` an incremental `make && ctest` re-runs the tests against the *previous* apps binaries. It does not affect CI, which always builds clean, but it silently gave me a wrong answer twice.

curtisc:
The two ppc64le failures (`SimpleEncRev53Raw32Signed`/`Unsigned`) turned out to be a latent bug in the VSX kernels that the new 32-bit tests are the first to reach, not a problem in the raw-reader fix — pushed f8e31f5 with the two-line correction.

**Root cause.** In `ojph_simd_vsx.h`, `vsx_i64x2_extend_low_i32x4` calls `vec_unpackl` and `vsx_i64x2_extend_high_i32x4` calls `vec_unpackh`, but the mapping is the other way around: `vec_unpackh` unpacks the half holding elements 0 and 1, `vec_unpackl` the half holding elements 2 and 3 (verified empirically with g++ on ppc64le — the compiler keeps that element numbering on little endian). Every si32 quad widened to si64 therefore came out `[x2, x3, x0, x1]`.

The only callers are the encode-side si32→si64 conversions in `ojph_colour_vsx.cpp` (`vsx_rev_convert`, `vsx_rev_convert_nlt_type3`, `vsx_rct_forward`), so a reversible encode of a component deep enough to need si64 lines wrote a codestream of the pair-swapped image. The decode side narrows with byte-indexed `vec_perm` shuffles and is unaffected — which is why the corruption survives the round trip and why nothing failed before: no existing test widens to si64. It also matches the CI numbers exactly: sample 0 decoded to sample 2's value (+131074 = two steps of the test ramp), with a `+2, +2, −2, −2` step pattern over each quad.

**Verification** (qemu-ppc64le, ubuntu 22.04, g++, same configure line as CI):
- Before the fix: a VSX-built `ojph_compress` produced a corrupt codestream that decoded identically wrong under both VSX and `OJPH_DISABLE_SIMD` builds of `ojph_expand`; a generic-built encode round-tripped bit-exact under both decoders — localizing the bug to the encode side.
- After the fix: VSX and `OJPH_DISABLE_SIMD` builds produce byte-identical codestreams, all four cross encode/decode pairings are bit-exact, and the full suite passes 94/94.

Happy to split this commit into its own PR if you'd rather keep the raw-reader fix isolated.

aous72:
Dear Curtis,

Thank you for this PR.  I did only a small touchup.

Kind regards,
Aous

PR #351, opened 2026-08-18
Add POWER8 support for the VSX kernels

runlevel5:
Add a `PPC_CPU_EXT_LEVEL_ARCH_2_07` level with matching HWCAP2 detection, and lower the transform, colour and codestream VSX kernels to an `-mcpu=power8` baseline so they are dispatched on POWER8 as well.

The block decoder keeps its POWER9 baseline and gate: it is load/permute bound and costs about 10% more when built for ISA 2.07, and it is never entered below POWER9, so the newer instructions there remain safe.


aous72:
Dear Trung,

Thank you for this useful PR.
I trust that the PowerPC community appreciate this addition.

Kind regards,
Aous

PR #356, opened 2026-08-23
Adding NLT type 2 and experimental type 4

aous72:
In this WIP PR, we are attempting to add NLT types 2, and an experimental type 4, which combines type 2 and 3.


aous72:
After studying the LUT, I discovered the following:
- Decoding requires a table look up, which is not a problem, as index samples are evenly spaced.
- For encoding, however, index samples are NOT equally spaces, and therefore a good way of doing this is to have an `inverse` table that, given the incoming/source value, gives us the LUT index.  The alternative is to do search, which a no-no.  For 16bit source samples (or half precision), the index of this inverse table is 16bit , which is manageable, and needs 128KB of memory. However, for 32-bit source samples, it is not manageable. It is conceivable to have a combination of a lookup table followed by search, but it is not ideal.  I suspect the guys who developed the LUT non-linearity did not consider 32-bit source data, or possibly 38 source data.

PR #358, opened 2026-08-25
fix(build): fix building AVX-512 runtime dispatch code with Clang/MSVC

### LINKED ISSUE #326: Building with ClangCL/MSVC for x86_64 fails with the default configuration
A lot of compiler errors of the same type within one file:
```
src\core\transform\ojph_transform_avx512.cpp(228,23): error : always_inline function '_mm512_set1_ps' requires target feature 'avx512f', but would be inlined into function 'avx512_irv_vert_step' that is compiled without support for 'avx512f'
    228 |       __m512 factor = _mm512_set1_ps(a);
        |                       ^
```
I used `OJPH_DISABLE_AVX512=ON` for now, it compiles fine. It's sorta ok as don't have AVX512 on my CPU, but I believe the library should compile cleanly to use runtime dispatching on CPUs with AVX512.

LLVM 22.1.8, I use `-flto -mavx2` among others, but this shouldn't have any impact on the mentioned file.

zlib-ng also compiles with AVX512 support by default, but doesn't have any build issues.

solbjorn:
On certain setups, for example, FullLTO, MSVC-style architecture / instruction set flags are not enough for Clang/MSVC (clang-cl). Pass more fine-grained GNU-style arch / IS flags to Clang under MSVC, which are still accepted directly and resolve such issues.

Closes #326

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -79,7 +79,7 @@ else()
       endif()
 
       # Set compilation flags
-      if (MSVC)
+      if (MSVC AND NOT CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
         set_source_files_properties(codestream/ojph_codestream_avx.cpp PROPERTIES COMPILE_FLAGS "/arch:AVX")
         set_source_files_properties(codestream/ojph_codestream_avx2.cpp PROPERTIES COMPILE_FLAGS "/arch:AVX2")
         set_source_files_properties(coding/ojph_block_decoder_avx2.cpp PROPERTIES COMPILE_FLAGS "/arch:AVX2")


aous72:
Hi Alexander,

Thank you for this PR; let's see if other people are fine with it.
It might be worthwhile for you to put a bug report to clang-cl people.

Kind regards,
Aous.

solbjorn:
@aous72, it's not a Clang/clang-cl issue -- with MSVC-style flags, you only have a basic IS flags like AVX-512 etc, there's no fine-grained switches. So most projects with runtime dispatch always pass GNU-style IS switches if the compiler is Clang, like I did in this PR.

aous72:
@solbjorn My view is that clang-cl claims to be compatible with cl.exe.  
Passing /arch:AVX512 to clang-cl does not produce the same effect of passing it to cl.exe.

PR #366, opened 2026-09-05
Decode HT codestreams with more than one quality layer

### LINKED ISSUE #89: ojph_expand error/warning behaviour with quality layer placeholder passes
While I'm not suggesting that using HT with quality layers is a great idea for mainstream use, it could be important in some special use cases and also when transcoding to/from J2K Part-1 codestreams which do commonly use quality layers.  I think the HTJ2K standard allows the representation of quality layers using the placeholder passes feature.  

Ideally, OpenJPH would be able to decode these files and if not, have a consistent response to files that are valid HTJ2K files yet not yet decodeable by OpenJPH.

ojph_expand has a different response to the attached input files with placeholder passes, depending on the input file

different errors and warnings are displayed
`A malformed codeblock that has more than one coding pass, but zero length for 2nd and potential 3rd pass.`
`Error decoding a codeblock`
`We do not support more than 3 coding passes; This codeblocks has 5 passes.`
`error in parsing a tile header; missing msbs are larger or equal to Kmax. The most likely cause is a corruption in the bitstream.`
`Segmentation Fault`

In the case of the pathological test with 65335 quality layers `ht_216p_65535_layers.j2c`, a gray image is decoded without any error or warning output. 

```
ojph_expand -i ht_216p_01_layers.j2c -o ht_216p_01_layers.j2c.tif
Elapsed time = 0.002735

ojph_expand -i ht_216p_02_layers.j2c -o ht_216p_02_layers.j2c.tif
ojph warning 0x00010001 at ojph_block_decoder_ssse3.cpp:1034: A malformed codeblock that has more than one coding pass, but zero length for 2nd and potential 3rd pass.
ojph error 0x000300A1 at ojph_codestream.cpp:4067: Error decoding a codeblock

ojph_expand -i ht_216p_07_layers.j2c -o ht_216p_07_layers.j2c.tif
ojph error 0x000300A1 at ojph_codestream.cpp:4067: Error decoding a codeblock

ojph_expand -i ht_216p_19_layers.j2c -o ht_216p_19_layers.j2c.tif
ojph warning 0x00010002 at ojph_block_decoder_ssse3.cpp:1042: We do not support more than 3 coding passes; This codeblocks has 5 passes.
ojph error 0x000300A1 at ojph_codestream.cpp:4067: Error decoding a codeblock

ojph_expand -i ht_216p_65535_layers.j2c -o ht_216p_65535_layers.j2c.tif
Elapsed time = 0.002206

ojph_expand -i ht_216p_01_layers.j2c -o ht_216p_01_layers.j2c.tif
Elapsed time = 0.002665

ojph_expand -i ht_2160p_01_layers.j2c -o ht_2160p_01_layers.j2c.tif
Elapsed time = 0.112171

ojph_expand -i ht_2160p_02_layers.j2c -o ht_2160p_02_layers.j2c.tif
ojph error 0x00030092 at ojph_codestream.cpp:1898: error in parsing a tile header; missing msbs are larger or equal to Kmax. The most likely cause is a corruption in the bitstream.

ojph_expand -i ht_2160p_10_layers.j2c -o ht_2160p_10_layers.j2c.tif
Segmentation fault

```
The input files were encoded with Kakadu v.8.2.1 using the following commands:
```
kdu_compress -i meridian_216p.tif -o ht_216p_01_layers.j2c Cmodes=HT -rate 1 Clayers=1 -no_info
kdu_compress -i meridian_216p.tif -o ht_216p_02_layers.j2c Cmodes=HT -rate 1 Clayers=2 -no_info
kdu_compress -i meridian_216p.tif -o ht_216p_07_layers.j2c Cmodes=HT -rate 1 Clayers=7 -no_info
kdu_compress -i meridian_216p.tif -o ht_216p_19_layers.j2c Cmodes=HT -rate 1 Clayers=19 -no_info
kdu_compress -i meridian_216p.tif -o ht_216p_65535_layers.j2c Cmodes=HT Clayers=65535 -no_info

kdu_compress -i 3840x2160_10bit_444_BT709_SDR.tif -o ht_2160p_01_layers.j2c Cmodes=HT -rate 1 Clayers=1 -no_info
kdu_compress -i 3840x2160_10bit_444_BT709_SDR.tif -o ht_2160p_02_layers.j2c Cmodes=HT -rate 1 Clayers=2 -no_info
kdu_compress -i 3840x2160_10bit_444_BT709_SDR.tif -o ht_2160p_10_layers.j2c Cmodes=HT -rate 1 Clayers=10 -no_info
```

The HJT2K J2C files generated by these commands are contained in the attached zip files:

[ht_2160p_01_layers.zip](https://github.com/aous72/OpenJPH/files/9105270/ht_2160p_01_layers.zip)
[ht_2160p_02_layers.zip](https://github.com/aous72/OpenJPH/files/9105271/ht_2160p_02_layers.zip)
[ht_2160p_10_layers.zip](https://github.com/aous72/OpenJPH/files/9105272/ht_2160p_10_layers.zip)
[ht_216p_01_layers.zip](https://github.com/aous72/OpenJPH/files/9105273/ht_216p_01_layers.zip)
[ht_216p_02_layers.zip](https://github.com/aous72/OpenJPH/files/9105274/ht_216p_02_layers.zip)
[ht_216p_07_layers.zip](https://github.com/aous72/OpenJPH/files/9105275/ht_216p_07_layers.zip)
[ht_216p_19_layers.zip](https://github.com/aous72/OpenJPH/files/9105276/ht_216p_19_layers.zip)
[ht_216p_65535_layers.zip](https://github.com/aous72/OpenJPH/files/9105277/ht_216p_65535_layers.zip)

jukjaa:
Fixes #89.

OpenJPH currently decodes only single-layer HT codestreams. `precinct::parse` reads one
packet per precinct and assumes it carries every contribution, so a codestream with
`SGcod.num_layers > 1` is rejected at header parse. This adds multi-layer decoding for all
five progression orders.

### What it does

A precinct's packets are now parsed one at a time with a quality-layer index, and the
sequencing loop gains a layer dimension — `l, r, c, p` for LRCP, `r, l, c, p` for RLCP, and
the layer read where the precinct is read for the three spatial orders. Per-code-block parse
state carries across layers, so a contribution's segments can arrive in different packets.

The segment rules follow T.814 Annex B: the general B.2 segment loop, zero-length HT sets,
cross-packet segment concatenation, `Z_blk` and `S_blk`, and the placeholder-run rule (P0
cannot be read off the first contribution, since a placeholder run may be split across
packets).

Two smaller things came out of the same work and are separate commits:

* Tag-tree storage per resolution was sized from the precinct/code-block *ratio*, which is
  generous for a precinct clipped by its tile. It is now sized from the code-blocks the
  precincts actually hold.
* A COC marker segment did not inherit SOP/EPH usage from the main COD (T.800 Table A.23),
  so a stream with COC plus SOP/EPH misparsed the SOP bytes as header bits.

### How it is split

The series is arranged so that review effort can concentrate on one commit:

| | commit | |
|---|---|---|
| 1 | `fix(decode): COC objects inherit SOP/EPH usage from the main COD` | independent bug fix |
| 2 | `refactor(decode): size a resolution's tag trees from its precincts` | |
| 3 | `refactor(decode): read one packet per precinct::parse call` | all the layer plumbing |
| 4 | `feat(decode): decode codestreams with more than one quality layer` | removes the guard |
| 5 | `test: cover decoding of multi-layer codestreams` | |

**Commits 1-3 keep the existing `num_layers != 1` guard, so they change nothing observable.**
Over a corpus of 151 codestreams — the 70 of `jp2k_test_codestreams`, the 73 of the
ISO/IEC 15444-4 conformance set, and the 8 attached to #89 — 86 decode on master, and at each
of commits 1-3 all 86 decode byte-identically, with nothing newly decoding and nothing lost.
Commit 4 removes the guard and turns the feature on. Each commit builds on its own.

### Verification

* **Byte-identical on everything that decodes today.** Across the 151-codestream corpus above,
  the full series decodes all 86 of master's outputs byte-for-byte, with 0 differing and 0
  regressions; 16 codestreams newly decode. Encoder output identical across five
  `ojph_compress` configurations.
* **Eleven conformance streams that master cannot decode now decode.** Master rejects every
  multi-layer stream — 28 of the 31 at the `num_layers` guard, and the three mixed-mode
  `ds0_hm_*` files earlier, at COD-SPcod — so the baseline is zero.

  | stream | layers | prog | against the ISO reference |
  |---|---|---|---|
  | `ds0_ht_02_b11` | 6 | LRCP | PAE 1, tolerance 1 |
  | `ds0_ht_02_b12` | 6 | LRCP | *no entry in the suite*; PAE 0 against the reference its `b11` sibling uses |
  | `ds0_ht_04_b11` | 20 | RLCP | PAE 2/2/2, tolerances 7/6/8 |
  | `ds0_ht_04_b12` | 20 | RLCP | PAE 2/2/2, tolerances 5/4/6 |
  | `ds0_ht_08_b11` | 30 | CPRL | PAE 32/30/33, tolerances 40/40/40 (suite decodes at `-reduce 1`) |
  | `ds0_ht_08_b15` | 30 | CPRL | PAE 0/1/0, tolerances 0/1/0 |
  | `ds0_ht_08_b16` | 30 | CPRL | PAE 0/0/0, exact |
  | `ds0_ht_10_b11` | 2 | LRCP | PAE 0/0/0, exact |
  | `ds0_ht_16_b11` | 3 | RLCP | PAE 0, exact |
  | `ds1_ht_01_b11` | 5 | LRCP | PAE 1, tolerance 1 |
  | `ds1_ht_01_b12` | 5 | LRCP | PAE 0, exact |

  Ten carry an entry in the suite's own cmake files and pass its PAE/MSE tolerances; four are
  exact. `ds0_ht_02_b12` has no entry there, so its comparison is mine rather than the
  suite's. `ds0_ht_16_b11` also decodes byte-identically to its single-layer sibling
  `ds0_ht_01_b11`. Five of the multi-layer codestreams attached to #89 newly decode as well.

  The three `ds0_ht_08` streams decode identically to OpenHTJ2K, on every sample. Worth a note
  if you reproduce it: they are 12-bit signed with three components, and `ojph_expand` has no
  sign-preserving output for that case — `.raw` preserves sign but is single-component, while
  PPM, TIFF and YUV clamp negatives to zero — so I compared sample values through the library
  API rather than through a written image.
* **A large multi-layer HT image from my own application** decodes to a byte-identical
  output before and after the change.
* **Builds warning-free** on MSVC x64, GCC 13.3, Clang 18.1.3, and Emscripten across all four
  variants your `emcc` workflow builds (SIMD/non-SIMD x Debug/Release). No new warning on any
  toolchain relative to master.
* **Fuzzed** with the libFuzzer targets from `OJPH_BUILD_FUZZER`, both `ojph_expand` and
  `ojph_compress`, 900s each against master as a baseline: 1.15M executions on expand and 48k
  on compress for this branch (1.17M / 66k on master), no crashes and no sanitizer reports on
  either. Peak RSS stayed at 468 MB. Coverage on this branch is higher than on master
  (cov 545 vs 513), which is the multi-layer path being reached.
* `ctest`: 97/97 -> 99/99 with the new tests.

### About the test

The multi-layer path cannot be covered the way the rest of the suite covers things. Every
other test generates its data with `ojph_compress`, but the encoder writes a single quality
layer and `param_cod` has no setter, so there is no round-trip to make.

The test therefore **generates its fixtures at run time** from a stream the library itself
encodes, by inserting packets that include no code-block to turn a one-layer codestream into
an N-layer one. Nothing binary is added to the repository. It covers the five progression
orders, two and three layers, and two forms of added layer — an empty packet, and a packet
that has a header but includes nothing, which makes the decoder walk a layer-1 packet header
and take the "already included" path.

The oracle is exact: the multi-layer stream must decode **bit-identically** to the one-layer
original, so no reference images are needed. Packet boundaries are located without a
packet-header parser, by encoding a second time with tile-part divisions, and the test asserts
its own layout model by re-assembling the one-layer tile and comparing it byte-for-byte with
what the encoder wrote.

**What the test does not cover**, stated plainly: a code-block whose passes are split *across*
layers, i.e. cross-packet segment concatenation and placeholder runs. That is not a matter of
effort — `ojph_compress` emits exactly one coding pass per code-block (the HT cleanup pass; HT
SigProp/MagRef encoding is not implemented), and a single pass cannot be split. Covering it
needs either an encoder that can write multiple passes or a committed multi-layer fixture.

I verified that path against the ISO/IEC 15444-4 conformance streams, which do exercise it,
but those are ISO-licensed and whether they belong in `jp2k_test_codestreams` is your call
rather than mine to make by putting them there. If you would like that coverage before this
merges, I am happy to prepare it — `ds0_ht_08_b11/_b15/_b16` (CPRL, 30 layers, COC on every
component, SOP+EPH) are good candidates, and they are what turned up the COC/SOP-EPH bug in
commit 1. Equally happy to leave it.

### Known limitations

* **The other twenty multi-layer conformance streams still do not decode**, each stopped by a
  feature OpenJPH does not have yet rather than by anything here — as they are for
  single-layer streams today: POC in the main header (`ds0_ht_03_b11/_b14`,
  `ds0_ht_15_b11/_b14`), POC in a tile (`ds0_ht_07_b11/_b15/_b16`), PPT
  (`ds1_ht_02_b11/_b12`), PPM (`ds1_ht_05_b11`), RGN (`ds0_ht_06_b11/_b15/_b18`),
  scalar-derived quantization in QCC (`ds0_ht_05_b11/_b12`, `ds1_ht_03_b11/_b12`), and mixed
  HT/T.800 mode (`ds0_hm_06_b11/_b18`, `ds0_hm_15_b8`, which master also rejects). The
  layer-aware sequencing here is a prerequisite for POC support, but POC itself is a separate
  feature.
* **At most 256 quality layers.** The inclusion tag tree stores the layer of first inclusion in
  one byte per node. Streams declaring more are rejected with a clear message rather than
  misparsed — `ht_216p_65535_layers.j2c` from #89 is one, and it is the only multi-layer
  codestream in that attachment set this PR does not decode. Lifting the limit means a wider
  node value for the inclusion tree specifically, since `ui8` is correct for the encoder's
  trees and the missing-MSB trees.
* **Memory.** Three costs, quoted in full:
  * `coded_cb_header` grows from 32 to 64 bytes per code-block, on **every** codestream,
    single-layer included, because the parse state has to survive between a code-block's
    packets.
  * On a codestream with more than one layer, each precinct gets its own tag-tree storage of
    16 tables, so its trees survive across its packets in every progression order. That is
    roughly 7 to 11 bytes per code-block at resolutions above 0, and 21 to 32 bytes at
    resolution 0, where a precinct holds only the LL band. Nothing on single-layer streams.
  * The shared precinct scratch grows from 4 tables to 16 — one buffer per codestream. For a
    precinct of 4x4 code-blocks that is 88 -> 352 bytes, and 1.33 -> 5.33 MiB in the largest
    case that can be constructed (precincts 2^15 samples wide with 64x64 code-blocks).

### DIFF
--- a/src/core/codestream/ojph_codeblock.h
+++ b/src/core/codestream/ojph_codeblock.h
@@ -118,6 +118,15 @@ namespace ojph {
       ui32 num_passes;       // number of passes to be decoded
       ui32 Kmax;
       ui32 missing_msbs;
+      // Parsing state, live while the packets of a precinct are read. It has
+      // to survive from one quality layer of a codeblock to the next, which
+      // is why it lives here and not in precinct::parse.
+      // included   : has contributed to an earlier quality layer
+      // in_layer   : contributes to the packet being read now
+      // Lblock_m3  : Lblock biased by 3, so zero means 3
+      ui8 included;
+      ui8 in_layer;
+      ui8 Lblock_m3;
       coded_lists *next_coded;
 
       static const int prefix_buf_size = 8;
--- a/src/core/codestream/ojph_codeblock.h
+++ b/src/core/codestream/ojph_codeblock.h
@@ -121,12 +121,34 @@ namespace ojph {
       // Parsing state, live while the packets of a precinct are read. It has
       // to survive from one quality layer of a codeblock to the next, which
       // is why it lives here and not in precinct::parse.
-      // included   : has contributed to an earlier quality layer
-      // in_layer   : contributes to the packet being read now
-      // Lblock_m3  : Lblock biased by 3, so zero means 3
+      // placeholder_passes : 3*P0 (T.814 B.1), or all ones while the
+      //                      placeholder run is still open
+      // pass_index         : index of the codeblock's next coding pass
+      // pkt_*              : the byte counts the packet in hand contributes,
+      //                      handed from its header to its data
+      // sets_seen          : HT sets seen so far
+      // decoded_set_skip   : S_skip of the set being decoded (T.814 B.3)
+      // included           : has contributed to an earlier quality layer
+      // in_layer           : contributes to the packet being read now
+      // has_cleanup        : holds a real HT cleanup segment
+      // starts_new_set     : this packet began a new HT set
+      // Lblock_m3          : Lblock biased by 3, so zero means 3
+      // set_passes         : coding passes gathered in the current HT set
+      // zero_bitplanes     : P, the missing MSBs the packet header codes
+      ui32 placeholder_passes;
+      ui32 pass_index;
+      ui32 pkt_pre_skip;
+      ui32 pkt_cleanup;
+      ui32 pkt_refine;
+      ui16 sets_seen;
+      ui16 decoded_set_skip;
       ui8 included;
       ui8 in_layer;
+      ui8 has_cleanup;
+      ui8 starts_new_set;
       ui8 Lblock_m3;
+      ui8 set_passes;
+      ui8 zero_bitplanes;
       coded_lists *next_coded;
 
       static const int prefix_buf_size = 8;
--- a/src/core/codestream/ojph_codestream_local.cpp
+++ b/src/core/codestream/ojph_codestream_local.cpp
@@ -796,10 +796,18 @@ namespace ojph {
           received_markers |= 1;
           ojph::param_cod c(&cod);
           int num_qlayers = c.get_num_layers();
-          if (num_qlayers != 1)
-            OJPH_ERROR(0x00030053, "The current implementation supports "
-              "1 quality layer only.  This codestream has %d quality layers",
-              num_qlayers);
+          // T.800 Table A.17: a codestream carries at least one quality
+          // layer. Without this a codestream that says zero would decode to
+          // an empty image rather than being reported.
+          if (num_qlayers < 1)
+            OJPH_ERROR(0x00030053, "This codestream says it has %d quality "
+              "layers; there has to be at least one.", num_qlayers);
+          // The inclusion tag tree stores the layer of first inclusion in
+          // one byte per node, so it cannot represent a layer index above
+          // 255.
+          if (num_qlayers > 256)
+            OJPH_ERROR(0x00030058, "This codestream has %d quality layers; "
+              "at most 256 are supported.", num_qlayers);
         }
         else if (marker_idx == 4)
         {
--- a/src/core/codestream/ojph_codestream_local.cpp
+++ b/src/core/codestream/ojph_codestream_local.cpp
@@ -43,6 +43,7 @@
 #include "ojph_params.h"
 #include "ojph_codestream_local.h"
 #include "ojph_tile.h"
+#include "ojph_precinct.h" // for precinct::num_parse_tag_trees
 
 #include "../transform/ojph_colour.h"
 #include "../transform/ojph_transform.h"
@@ -211,8 +212,11 @@ namespace ojph {
       // We need 4 such tables. These tables store
       // 1. missing msbs and 2. their flags,
       // 3. number of layers and 4. their flags
+      // A precinct is parsed with one set of these per subband, so the
+      // buffer holds precinct::num_parse_tag_trees of them.
       precinct_scratch_needed_bytes =
-        4 * ((max_ratio * max_ratio * 4 + 2) / 3);
+        precinct::num_parse_tag_trees
+        * ((max_ratio * max_ratio * 4 + 2) / 3);
 
       allocator->pre_alloc_obj<ui8>(precinct_scratch_needed_bytes);
     }
--- a/src/core/codestream/ojph_codestream_local.cpp
+++ b/src/core/codestream/ojph_codestream_local.cpp
@@ -183,40 +183,38 @@ namespace ojph {
         allocator->pre_alloc_obj<param_tlm::Ttlm_Ptlm_pair>(num_tileparts);
 
       //precinct scratch buffer
-      // The precinct scratch is shared by all components, but each component
-      // may override the codeblock/precinct geometry via a COC marker.  The
-      // per-component tag-tree storage (resolution.cpp) is derived from that
-      // component's effective params, so size the shared buffer from the
-      // largest ratio across every component (the main COD and all COC
-      // overrides).  Sizing from the COD alone under-reserves the buffer for
-      // any component whose COC declares a smaller codeblock than the COD.
-      size ratio;
+      // Shared by all components, so sized from the largest codeblock count
+      // any of them reaches; a COC may declare a smaller codeblock than the
+      // COD, and sizing from the COD alone would under-reserve. Bounded by
+      // the component extent, as resolution::pre_alloc is.
+      size max_cbs(1, 1);
       for (ui32 c = 0; c < num_comps; ++c)
       {
         const param_cod* cdp = cod.get_coc(c);
         ui32 num_decomps = cdp->get_num_decompositions();
         size log_cb = cdp->get_log_block_dims();
+        ui32 shift_w = ojph_min(log_cb.w, (ui32)31);
+        ui32 shift_h = ojph_min(log_cb.h, (ui32)31);
         for (ui32 r = 0; r <= num_decomps; ++r)
         {
           size log_PP = cdp->get_log_precinct_size(r);
-          ratio.w = ojph_max(ratio.w, log_PP.w - ojph_min(log_cb.w, log_PP.w));
-          ratio.h = ojph_max(ratio.h, log_PP.h - ojph_min(log_cb.h, log_PP.h));
+          ui32 by_ratio_w = 1u << (log_PP.w - ojph_min(log_cb.w, log_PP.w));
+          ui32 by_ratio_h = 1u << (log_PP.h - ojph_min(log_cb.h, log_PP.h));
+          ui32 ds = num_decomps - r;
+          ui32 res_w = siz.get_recon_width(c) >> ojph_min(ds, (ui32)31);
+          ui32 res_h = siz.get_recon_height(c) >> ojph_min(ds, (ui32)31);
+          max_cbs.w = ojph_max(max_cbs.w,
+            ojph_min(by_ratio_w, (res_w >> shift_w) + 2));
+          max_cbs.h = ojph_max(max_cbs.h,
+            ojph_min(by_ratio_h, (res_h >> shift_h) + 2));
         }
       }
-      ui32 max_ratio = ojph_max(ratio.w, ratio.h);
-      max_ratio = 1 << max_ratio;
-      // assuming that we have a hierarchy of n levels.
-      // This needs 4/3 times the area, rounded up
-      // (rounding up leaves one extra entry).
-      // This exta entry is necessary
-      // We need 4 such tables. These tables store
-      // 1. missing msbs and 2. their flags,
-      // 3. number of layers and 4. their flags
-      // A precinct is parsed with one set of these per subband, so the
-      // buffer holds precinct::num_parse_tag_trees of them.
+      // Missing msbs, the layer of first inclusion, and a flag tree for
+      // each: one set per subband, for num_parse_tag_trees in all.
       precinct_scratch_needed_bytes =
-        precinct::num_parse_tag_trees
-        * ((max_ratio * max_ratio * 4 + 2) / 3);
+        (size_t)precinct::num_parse_tag_trees
+        * (size_t)precinct::num_tag_tree_bytes(
+                    precinct::num_tag_tree_levels(max_cbs));
 
       allocator->pre_alloc_obj<ui8>(precinct_scratch_needed_bytes);
     }
--- a/src/core/codestream/ojph_params_local.h
+++ b/src/core/codestream/ojph_params_local.h
@@ -559,6 +559,12 @@ namespace ojph {
       }
 
       ////////////////////////////////////////
+      ui32 get_num_layers() const
+      {
+        return SGCod.num_layers;
+      }
+
+      ////////////////////////////////////////
       // SOP and EPH usage is signalled only in Scod of the COD marker
       // segment; Scoc carries just the precinct bit (T.800 Table A.23). A
       // component that has a COC therefore inherits these two flags from
--- a/src/core/codestream/ojph_params_local.h
+++ b/src/core/codestream/ojph_params_local.h
@@ -559,11 +559,15 @@ namespace ojph {
       }
 
       ////////////////////////////////////////
+      // SOP and EPH usage is signalled only in Scod of the COD marker
+      // segment; Scoc carries just the precinct bit (T.800 Table A.23). A
+      // component that has a COC therefore inherits these two flags from
+      // the COD it overrides, rather than reading them as absent.
       bool packets_may_use_sop() const
       {
         if (type == COD_MAIN || type == COD_TILE)
           return (Scod & 2) == 2;
-        return false;
+        return top_cod != NULL ? top_cod->packets_may_use_sop() : false;
       }
 
       ////////////////////////////////////////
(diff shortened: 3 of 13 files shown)


aous72:
Dear Jukka,

Thank you for this PR.

I did not consider adding this feature because I assumed that it is of little utility for most of the community.

As you know, with HTJ2K each quality layer subsumes (include all information) of earlier layers; therefore, for normal decoding, you only need to decode the latest layer (the one with highest index).  Normal HTJ2K encoders would usually generate only one quality layer. Generating many quality layers produces a large file.
Kakadu can generate multiple quality layers, if you supply the correct command line options.

Supporting multiple quality layers also adds complexity to the code and perhaps needed memory.

So, do you have an application for this? I would be very much interested to know.

Kind regards,
Aous.

jukjaa:
Dear Aous,

Thank you for the quick reply; it is a fair question.

I agree with the property, and I think the standard states it at one granularity: the HT set. Clause 7.1.1 (NOTE 1) has a decoder process only the Z_blk passes of one cleanup segment and the refinement segment following it, and 8.3 adds that it need not decode more than one HT set per code-block. The patch does that — one set per code-block, the last non-empty one, earlier sets skipped rather than decoded.

What does not carry over is the step from sets to layers. A set subsumes those before it, but a layer is only a partition of a code-block's coding passes, and a set need not sit inside one layer. A SINGLEHT codestream (Ccap15 bit 13 = 0, A.3.3) has at most one set per code-block, so extra layers leave nothing to subsume: they share out the placeholder run and that set's cleanup, SigProp and MagRef passes. The top layer then usually holds a refinement segment unusable without the cleanup below it, and the placeholder passes, carrying no bytes, are what fix P0.

In every stream I can find this is the normal case, not the exception. Every HT codestream in the 15444-4 conformance data I have (the copy shipping with OpenHTJ2K) declares SINGLEHT, including all 31 with more than one layer, as do the eight Kakadu streams on #89. On `ds0_ht_08_b11.j2k` (30 layers, CPRL), which an independent parser accounts for to the byte — 2,444,802 over 5,130 packets, no rule violations — 2,520 of the 2,523 code-blocks contribute to more than one layer (up to 17), every one has exactly one cleanup segment, 1,912 (76%) reach a highest layer carrying only refinement passes, and 2,517 have their placeholder run below the cleanup.

One mechanical point holds even for MULTIHT: the last set cannot be reached by seeking to the last layer, since the inclusion tag tree codes the layer of first inclusion and Lblock and the missing-MSB trees carry forward. "Decode only the latest set" is realisable only as "do not entropy-decode superseded sets", inside a parser that has followed every layer — which is what this PR is.

On utility I would put the conformance suite ahead of my own files. Of my 40 HT streams, 31 declare more than one layer and none decode today: 28 stop at the COD layer guard, the three mixed-mode `ds0_hm_*` files earlier, at SPcod. This PR gets eleven decoding: ten against the suite's own reference images and tolerances, and `ds0_ht_02_b12`, which the suite has no entry for, exactly against the reference its sibling uses. The other twenty stop at features OpenJPH lacks — POC, PPT, PPM, RGN, scalar-derived QCC, mixed mode — separate work. Per-stream detail will go in the description. One reproduction note: `ds0_ht_08_b11/_b15/_b16` come out identical to OpenHTJ2K, but they are 12-bit signed with three components, and `ojph_expand` has no sign-preserving output for that case (`.raw` is single-component, and PPM, TIFF and YUV clamp negatives to zero), so I compared sample values through the library API. Multi-layer HT also arises the way the B.1 NOTE describes: placeholder passes preserving the quality layer boundaries of a Part 1 codestream that has since been transcoded — the same shape as the file above.

My own case is digital pathology. Whole-slide images reach us as multi-layer HTJ2K from a third-party encoder we do not control, so reading them as they are is our only option.

Your three practical points I agree with entirely: most encoders write one layer, OpenJPH's included; many layers inflate a file with packet headers; and the cost is real. I would rather quote it than reassure you, and since the description understates it I will correct it there too:

* `coded_cb_header`: 32 to 64 bytes per code-block, on every codestream — parse state must survive between a code-block's packets.
* With more than one layer, each precinct gets its own 16 tag-tree tables so its trees survive its packets in every progression order: about 7 to 11 bytes per code-block above resolution 0, more at resolution 0 (LL band only), nothing on single-layer streams.
* Shared scratch: 4 tables to 16 — 88 to 352 bytes for a 4x4-code-block precinct, 1.33 to 5.33 MiB in the largest case I can construct (precincts 2^15 samples wide, 64x64 code-blocks). The description's "22 -> 88 bytes" is wrong; the figures are 88 to 352.
* The description also claims eight streams newly decode; that was against an earlier version of my own work, not master. Against master it is eleven.

Commits 1-3 keep the single-layer guard and change nothing observable; only commit 4 removes it.

If you would prefer a smaller first step, I am glad to reduce the scope — leaving LRCP and RLCP rejected and supporting only the orders that keep a precinct's packets together, which drops the per-precinct tag-tree storage and the layer dimension in the sequencing loop, roughly half the change. And if you would rather not carry the feature at all, I understand completely; it has been useful to me either way.

Kind regards,
Jukka

michaeldsmith:
I think correctly decoding the HTJ2K conformance codestream files is a good goal.  That 65535 layer file I created for issue #89 is of course an extreme case.  If it really slows down the OpenJPH code to support decoding multiple layers, then perhaps a adding a compile time conformance mode might be another option?  One of the stated goals of the HTJ2K project was to allow roundtrip transcoding  J2K1 -> HTJ2K -> J2K1.  Another use case that was anticipated was fast encoding of huge images using HTJ2K including reasonable quality layer boundaries via placeholder passes, those HTJ2K files could then, at a later time, be transcoded with the good quality layer boundaries back to J2K1 for a low-bitrate quality-layer partial decoding or a remote-network-access decoding paradigm, for example JPIP.

aous72:
Dear @jukjaa,

Thank you for your detailed explanation.
I did not expect that. 
It is not clear to me why the third part choose this option. Perhaps, they are confused, but I can be wrong.
You can check if they are using Kakadu, a comment in the codestream would say that, unless they chose to remove it.
Generating multi quality layer HTJ2K requires more work, and there is a question of how to divide the layers.
You do not need to tell me; you can check for your own knowledge.

I will think of how to add this -- you have done a lot of work already.

Kind regards,
Aous 

Dear @michaeldsmith,

Thank you for your explanation and I agree that being able to decode conformance codestreams is a good idea.
I wanted to do this, but never had the time.

Kind regards,
Aous

jukjaa:
Dear Aous,

I pushed a fix for the out-of-memory CIFuzz found: the tag tree reservation was sized from the declared precinct-to-code-block ratio, so a header naming a large precinct over a small image reserved trees for code-blocks that cannot exist.

CIFuzz now hits a different crash — a null dereference from the sequencing loop in `tile::parse_tile_header`, through `tile_comp::get_top_left_precinct`. The input is 216 bytes, base64:

```
Oy3/////////////c3Nzc3Nzc3Nzc3P/AAATAAAAAAAA/////ywA////+fgBAAABIAD/T/9RAClAAAAAAvEAAAECAAABQAAAAIoAAAHz/QcAAAAAAAAAAAAAAAEfAQEAAAD5/wAAAAAg/1IAEvkEAAEABQAEQAFkRGRkZDg4NzcyNOMx/1wAEyBACgAAAAAAAAAAAAAAAC///woA////kP///68IAAAv/5AACgAAACABANt5YP+ThoaGenl5eXl5eXGGhoaGhoaGhieGhoaGTExMTEyGCSjL
```

One dead end, in case it saves you time: making the spatial orders honour `skipped_res_for_read`, as LRCP and RLCP already do, is not it — that changes the output of valid codestreams decoded at reduced resolution.

Happy to dig further if it would help.

Kind regards,
Jukka

jukjaa:
Dear Aous,

A short follow-up on your question about the encoder. I asked the vendor, and your reading was right: the multiple layers are not deliberate. They are a bit-rate ladder default inherited from their JPEG 2000 Part 1 pipeline, applied unchanged when they moved to HTJ2K. Nothing in their workflow consumes the layers, and they are moving new output to a single layer.

The files already produced stay as they are and will not be re-encoded, so reading them remains necessary. That is the application, such as it is.

Kind regards,
Jukka

aous72:
Love your comment 😄

PR #369, opened 2026-09-15
README.md: use HTTPS for PDF links

PeterSchuebel:
Links to papers should use HTTPS links, not HTTP


aous72:
Good idea.  Thank you Peter.

Issue #375, opened 2026-09-21
Type-2 and Type-4 NLT encoding - 2 issues


michaeldsmith:
In OpenJPH 0.32.0 the encoder's inverse-LUT for NLT types 2 and 4 has two problems.

1) Its first and last entries have the wrong values, so samples near the two ends of the LUT's range are encoded incorrectly.

2) It is a uniformly sampled approximation of the true inverse, which puts a floor under the lossy coding error.

Neither error shrinks as qstep gets smaller. Below is some background text that helps me explain the problems.  Additional details about the problems are at the end of the text.

The Type-2 LUT and Type-4 LUT assume a uniform input sampling. This means that only the output values need to be stored to represent the LUT, in addition to the input range [Dmin, Dmax] over which the input is uniformly sampled. Values that fall between two entries are linearly interpolated.

Thus even though the LUT is defined with only 1 set of output values, it can be conceptually thought of as pairs of (input, output) values, like this, where delta = (Dmax - Dmin) / (Npoints - 1) and Npoints is the number of entries in the LUT:

| coded-value (NLT input) | sample-value (NLT output) |
|---|---|
| Dmin | LUT[0] |
| Dmin + 1 x delta | LUT[1] |
| Dmin + 2 x delta | LUT[2] |
| ... | ... |
| Dmax - 2 x delta | LUT[Npoints-3] |
| Dmax - 1 x delta | LUT[Npoints-2] |
| Dmax | LUT[Npoints-1] |

The typical JPEG 2000 decoding process produces coded-values. The non-linear transform that follows the decoding process transforms the coded-values to sample-values using the non-linear transform described by the LUT.

The encoder does the inverse flow. Its non-linear transform consumes sample-values and produces coded-values. The coded-values are then sent to the regular JPEG 2000 encoding process.  Therefore the encoder must apply the inverse of the non-linear transform that is applied in the decoding flow.

Conceptually, the inverse-LUT can be described by the same values as the LUT but with the pairs of input <-> output values swapped, like this:

| sample-value (inverse-NLT input) | coded-value (inverse-NLT output) |
|---|---|
| LUT[0] | Dmin |
| LUT[1] | Dmin + 1 x delta |
| LUT[2] | Dmin + 2 x delta |
| ... | ... |
| LUT[Npoints-3] | Dmax - 2 x delta |
| LUT[Npoints-2] | Dmax - 1 x delta |
| LUT[Npoints-1] | Dmax |

[...]

PR #376, opened 2026-09-21
Fix ends of the encoding LUT of LUT-style NLTs

palemieux:
nlt_rec::prepare_for_encoding() builds the table that inverts the LUT of a type 2 or type 4 NLT. Each entry of that table is a position in the domain the LUT is indexed over, but the first and the last entries were set to ft_min and ft_max, which are the values of the first and last entries of the LUT, i.e. positions in its range. The two only coincide when the LUT entries span the same interval as the LUT is indexed over.

When they do not, samples that fall in the first or the last cell of the encoding table are interpolated towards the wrong end point and are encoded far from where they belong. For example, with a LUT indexed over [0, 2^32) whose entries span [0.22, 0.78] of it, samples within about 9% of the largest value (65504 in a float32 transfer function LUT) came back from a round trip with an error of up to 99%.

Set the end points to fd_min and fd_max, the positions of the first and last entries of the LUT, in the 1, 2 and 4 byte code paths.

Add a test that round trips a ramp through a type 2 and a type 4 LUT whose entries only cover the middle half of the range they are indexed over. It fails without this change (error of 2^29 on the last sample) and passes with it.


michaeldsmith:
It would be nice to also fix the comment for members `ft_min` and `ft_max` on line 893 OpenJPH\src\core\codestream\ojph_params_local.h
 
`float ft_min, ft_max;  // float d_min and d_max`

to something like

`float ft_min, ft_max;  // float first and last LUT points`

michaeldsmith:
Also, "Closes #375" is not completely accurate, since that included a description of two problems 1) The end-point bug this PR addresses and 2) the error-floor due to encoding using a uniform inverse-LUT.  

I don't think we want to have merging this PR auto-close issue #375 since the problem #2 (error-floor) in that issue is not addressed by this PR.

palemieux:
> Also, "Closes https://github.com/aous72/OpenJPH/issues/375" is not completely accurate, since that included a description of two problems 1) The end-point bug this PR addresses and 2) the error-floor due to encoding using a uniform inverse-LUT.

Good call. Can you create two separate issues?

Issue #377, opened 2026-09-21
Type-2 and Type-4 NLT - uniform sampled inverse-lut encoding causes error floor


michaeldsmith:
issue #375 described two problems in the inverse-LUT encoding:

1) a start and ending inverse-LUT value issue
2) an error-floor that caused by the uniformly sampled inverse-LUT implementation, which prevents the lossy coding error from going to zero as the qstep value is descreased

issue #375 has a good background and detailed description of why these issues happen and how to avoid them.  

Problem 1 was addressed in PR #376.

This issue was added to track Problem 2, the error floor due to uniformly sampled inverse-LUT used during the encoding.

aous72:
Thank you for raising issue #375, and for all the details there.

Issue 1 is a bug, but I am not sure if anyone was affected by it.
Issue 2 is more serious as it is core to the algorithm I am implementing.  These are my thoughts:
1. The current implementation has an error floor as detailed in #375; however, this is the most performant implementation.
2. While it is true that at higher quality, the quality of the current implementation saturates, the real question is whether we need higher precision here, because we are using LUT with lossy compression.  It is enough for the error due to using the LUT to be smaller than the smallest quantization error we need to use.
3. I considered a different design with higher computational complexity and higher accuracy in my earlier discussion with @palemieux, described below. 

An alternate method which requires one step search:
1. For encoding, we sample the encoding LUT densely enough such that each edge of the decoding LUT is visited at least once.
2. During encoding, a sample being encoded is either before or after the node (or kink).
3. If it is before the kink, then we use the lower point and the kink point. If it is after then kink, then we use the kink and the following point.
4. For SIMD implementation, we need to do 3 lookups (instead of the current 2), and we need a division (which is currently not needed); or we pre-compute slopes, and in this case, we need 5 lookups (no division).  The result is higher complexity with no advantage where the LUT is really needed.
5. This method should be accurate to floating-point rounding error.

The caveat is that if the non-linearity has close to a horizontal line (shallow slopes), the lookup table will become too big, or we need to use multiple-step search (SIMD becomes impossible).

Kind regards,
Aous

PR #378, opened 2026-09-22
Fix comment on ft_min and ft_max in nlt_rec

### LINKED ISSUE #375: Type-2 and Type-4 NLT encoding - 2 issues

michaeldsmith:
- fixes comment associated with the bug fixed in PR #376, described in issue #375
- this PR closes #375 
- the second problem described in #375 now has a unique issue #377 to track it.


aous72:
Thank you @michaeldsmith.

